PR #15, opened 2020-07-28
Added execute liberty dev task, expand/collapse all in tree view, and validation for custom start parameters

kathrynkodama:
- Expand/Collapse for Tree View
image

- Execute Liberty Dev Task
image

    If a user selects one task and then the "Execute Liberty Dev Task" button, they will be asked to confirm their selection
    image

    If a user selects the "Execute Liberty Dev Task" button while multiple or no tasks are selected, they will be asked to choose a task and then a project to execute on.
    image
    image

- Improved the view of the Custom Start dialog and added validation (input must start with `-`) to match the VS Code extension
image

- Added additional logging debug messages


src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -60,6 +69,8 @@ public static Tree buildTree(Project project, Color backgroundColor) {
         String projectName = null;
```
ericglau, line 69: Since `projectName` is used inside each of the `for (PsiFile file : mavenBuildFiles) {` and `for (PsiFile file : gradleBuildFiles) {` loops and changes between iterations, it's better to declare the variable within the scope of each of those loops instead of here.  Otherwise it gives the impression that this variable is needed outside of the loops.

PR #16, opened 2020-07-29
Terminal issue fix, collapse/expand fix, execute task fix

kathrynkodama:
- Updated vendor name to `Open-Liberty` in `plugin.xml`
- Fix in [getTerminalWidget](https://github.com/OpenLiberty/open-liberty-tools-intellij/compare/master...kathrynkodama:dev?expand=1#diff-48f1ecd020366bdb7adaa61fefdd87b1R64) to address Linux issue where the Linux machine could not connect to the terminal
- Modified the order of the `Execute Liberty Dev Task` so that user is prompted to select a project first and then a task to execute.  Tasks available to select are now filtered on project type
image
image

- Fix for the collapse/expand all buttons in the dashboard.  Previously if users hit the refresh button the collapse and expand all buttons stopped working.  Now they are linked to the updated Tree object.

Signed-off-by: Kathryn Kodama <kathryn.s.kodama@gmail.com>

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyToolbarActionGroup.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyToolbarActionGroup.java
@@ -1,0 +1,49 @@
+package io.openliberty.tools.intellij.actions;
+
+import com.intellij.openapi.actionSystem.ActionManager;
+import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.actionSystem.DefaultActionGroup;
+import com.intellij.ui.treeStructure.Tree;
+import com.intellij.ui.treeStructure.actions.CollapseAllAction;
+import com.intellij.ui.treeStructure.actions.ExpandAllAction;
+import io.openliberty.tools.intellij.util.Constants;
+
+public class LibertyToolbarActionGroup extends DefaultActionGroup {
+
+    CollapseAllAction collapseAction;
+    ExpandAllAction expandAction;
+    Tree tree;
+    public LibertyToolbarActionGroup(Tree tree) {
+        super("LibertyActionToolBar", false);
+        this.tree = tree;
+        final ActionManager actionManager = ActionManager.getInstance();
+        add(actionManager.getAction("io.openliberty.tools.intellij.actions.RefreshLibertyToolbar"));
+        addSeparator();
+        add(actionManager.getAction("io.openliberty.tools.intellij.actions.ExecuteLibertyDevTask"));
+        addSeparator();
+
+        this.collapseAction = new CollapseAllAction(tree);
+        this.expandAction = new ExpandAllAction(tree);
+        add(this.collapseAction);
+        add(this.expandAction);
+    }
+
+    @Override
+    public void update(AnActionEvent event) {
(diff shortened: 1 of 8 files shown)


src/main/java/io/openliberty/tools/intellij/actions/LibertyToolbarActionGroup.java
```diff
@@ -0,0 +1,49 @@
...
+import com.intellij.ui.treeStructure.Tree;
+import com.intellij.ui.treeStructure.actions.CollapseAllAction;
+import com.intellij.ui.treeStructure.actions.ExpandAllAction;
+import io.openliberty.tools.intellij.util.Constants;
+
+public class LibertyToolbarActionGroup extends DefaultActionGroup {
+
+    CollapseAllAction collapseAction;
+    ExpandAllAction expandAction;
+    Tree tree;
```
ericglau, line 15: Should these be marked as private?

src/main/java/io/openliberty/tools/intellij/util/Constants.java
```diff
@@ -50,16 +51,33 @@
...
+        MAVEN_ACTIONS_MAP.put(VIEW_UNIT_TEST_REPORT, VIEW_UNIT_TEST_REPORT_ACTION_ID);
+        MAVEN_ACTIONS_MAP.put(VIEW_INTEGRATION_TEST_REPORT, VIEW_INTEGRATION_TEST_REPORT_ACTION_ID);
+        return MAVEN_ACTIONS_MAP;
+    }
+
+    public static HashMap<String, String> getGradleMap() {
+        HashMap<String, String> GRADLE_ACTIONS_MAP = CORE_ACTIONS_MAP;
+        GRADLE_ACTIONS_MAP.put(VIEW_GRADLE_TEST_REPORT, VIEW_GRADLE_TEST_REPORT_ACTION_ID);
+        return GRADLE_ACTIONS_MAP;
+    }
```
ericglau, line 82: Note that doing like `HashMap<String, String> a = b;` does not make a copy of the map.  So when you are doing `put` afterwards in these 3 methods, you are actually modifying the original map.  To make a copy, you could do `a = new HashMap<String, String>()` then `a.putAll(b)`

ericglau: Also things like FULL_ACTIONS_MAP in the method is a local variable and not a constant, so convention should be ordinary camel case.

src/main/java/io/openliberty/tools/intellij/util/Constants.java
```diff
@@ -50,16 +51,33 @@
     public static final String VIEW_GRADLE_CONFIG_ACTION_ID = "io.openliberty.tools.intellij.actions.ViewGradleConfig";
     public static final String VIEW_EFFECTIVE_POM_ACTION_ID = "io.openliberty.tools.intellij.actions.ViewEffectivePom";
 
-    public static final Map<String, String> ACTIONS_MAP = new HashMap<String, String>() {
+    public static HashMap<String, String> CORE_ACTIONS_MAP = new HashMap<String, String>() {
```
ericglau, line 54: Should this be private if this constant is only used within this class?

PR #111, opened 2022-10-19
Support for the first three diagnostics for Jakarta LS

turkeylurkey:
I tried to leave the code as similar as possible to the JDT based code rather than optimize and rewrite. This is so that future support can be as simple as possible. 

Also, I should update this PR with the changes required by the Jakarta language server client API update.


turkeylurkey:
"Add AnnotationDiagnosticsCollector for jakarta.annotation.* annotations." adds AnnotationConstants, AnnotationDiagnosticsCollector, Messages and org.eclipse.lsp4jakarta.jdt.core.messages.messages.properties.

kathrynkodama (approved):

PR #131, opened 2022-11-09
Bump intellij version

evie-lau:
To build the plugin successfully from scratch, needed to use Java 17 for building `lsp4jakarta`.
Then, to build the plugin itself with Java 17, had to bump a few versions for the IntelliJ platform.
Added GHA build to test.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,15 +14,25 @@ jobs:
         os: [ubuntu-latest, windows-latest]
 
     steps:
-      - uses: actions/checkout@v3
+      - name: Checkout liberty-tools-intellij
+        uses: actions/checkout@v3
+        with:
+          path: liberty-tools-intellij
+      - name: Checkout lsp4jakarta
+        uses: actions/checkout@v3
+        with:
+          repository: eclipse/lsp4jakarta
+          path: lsp4jakarta
       - name: Set up Java
         uses: actions/setup-java@v3
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
           cache: 'gradle'
       - name: Build Liberty-Tools-Intellij
-        run: gradle buildPlugin
+        run: |
+          gradle buildJakarta
+          gradle buildPlugin
       - name: Archive artifacts
         if: runner.os == 'Linux'
         uses: actions/upload-artifact@v3
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -8,10 +8,7 @@ on:
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-    strategy:
-      matrix:
-        os: [ubuntu-latest, windows-latest]
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout liberty-tools-intellij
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -0,0 +1,57 @@
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+      - name: Build Liberty-Tools-Intellij
+        working-directory: ./liberty-tools-intellij
+        run: |
+          gradle buildJakarta
+          gradle buildPlugin
```
kathrynkodama, line 49: Should these use the project's Gradle wrapper?

evie-lau: fixed!

evie-lau:
While these changes allow the project to build successfully with Java17, we will need some testing in various environments to see if we are still able to run successfully with `runIde`.

build.gradle
```diff
@@ -19,6 +20,9 @@ repositories {
     maven {
         url 'https://repo.eclipse.org/content/repositories/lemminx-releases'
     }
+    maven {
```
turkeylurkey, line 23: Usually we only use sonatype during development. Is there a dependency we can only get there or is this a line we'll remove later like `mavenLocal()`?

build.gradle
```diff
@@ -74,14 +78,39 @@ runIde {
     jvmArgs '--add-exports', 'java.base/jdk.internal.vm=ALL-UNNAMED'
 }
 
+
+task buildJakartaCore(type: Exec) {
```
turkeylurkey, line 82: My understanding is lsp4jakarta is a separate project. Do we need these build instructions here?

build.gradle
evie-lau: LCLS/lemminx-liberty jars weren't available in the other maven repos, and it failed to download/start without the sonatype repo. I'll check with Matt to see if it's published anywhere else.

build.gradle
evie-lau: Our ls-integration wants to include lsp4jakarta, which I believe is not published anywhere yet, so we have to clone and build/install it locally. I figured `build.gradle` was the most convenient place to put it for developing, should it be somewhere else?

build.gradle
turkeylurkey: Ok, I assume open source users can build the dependencies locally if they can't access sonatype.

build.gradle
turkeylurkey: There is a build script for lsp4jakarta in the root directory of that project called `buildAll.sh`. The code in liberty-tools-intellij should just be for building itself. The interface to other projects is contained in the `dependencies {}` section. 

If one wanted to build both lsp4jakarta and liberty-tools-intellij in one go perhaps a local script in the directory containing these two projects.

TrevCraw: Now that LSP4Jakarta has snapshots available, we should pull from the Eclipse repo. We will address this with future changes.

turkeylurkey: Will review in future.

turkeylurkey (approved):

Issue #154, opened 2022-11-23
[Windows] When the path for maven has spaces, we get an issue running liberty on Intellij


fabio-franco:
When running liberty on Windows and the path for maven has spaces, we got an issue:

```
PS C:\Users\TREVORCrawford\Git\guide-rest-intro-2> cd "C:/Users/TREVORCrawford/Git/guide-rest-intro-2/finish"
PS C:\Users\TREVORCrawford\Git\guide-rest-intro-2\finish> cmd /K C:\Program Files\JetBrains\IntelliJ IDEA Community Edition 2022.2.3\plugins\maven\lib\maven3\bin\mvn io.openliberty.tools:liberty-maven-plugin:dev
'C:\Program' is not recognized as an internal or external command,
operable program or batch file.
```

That is potentially happening for Gradle as well. Need to check.

PR #158, opened 2022-11-24
Including double quotes to escape spaces in paths

fabio-franco:
This is a fix for the issue: https://github.com/OpenLiberty/liberty-tools-intellij/issues/154

Including double quotes, we allow spaces in the path for Gradle/Maven executables.

This was tested over Mac and Windows environments.


kathrynkodama (approved):

mrglavas (approved):
Looks good to me. If it's possible for the file path to contain double quotes or other characters that are special to the command line interpreter, it may be worth opening another GitHub issue to track potential problems with those characters.

Issue #127, opened 2022-11-09
Action executed from shift + shift window may use context from cancelled action.


mrglavas:
Steps to reproduce:

1) Select "Start dev mode..." from the project window.

![image](https://user-images.githubusercontent.com/890521/200861911-2935c4f6-ce3d-41ce-bf63-2e97ea32d1e5.png)

2) Select "Cancel" from the pop up dialog.

![image](https://user-images.githubusercontent.com/890521/200862299-54e1086c-9b36-4047-bd17-42d0f9c0047b.png)

3) Press shift + shift and select an action to run.

![image](https://user-images.githubusercontent.com/890521/200862793-6b8bfe2a-f9ce-4e4e-aa7a-2ef65bf7a5be.png)

Instead of prompting the user to select a project for the action, the action runs with the project that had been selected for the previously cancelled action.

mrglavas:
Might be related to: https://github.com/OpenLiberty/liberty-tools-intellij/issues/126.

kathrynkodama:
Just adding that I see the same behaviour @TrevCraw

PR #166, opened 2022-11-24
Fix for shift-shift project selection

fabio-franco:
This is a fix for the issue: https://github.com/OpenLiberty/liberty-tools-intellij/issues/127

Tests were done on Intellij over Mac.


src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -65,13 +65,16 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
-        if (buildFile == null) {
+
+        boolean isActionFromShiftShift = "GoToAction".equalsIgnoreCase(e.getPlace());
+
+        if (isActionFromShiftShift || buildFile == null) {
             buildFile = (VirtualFile) e.getDataContext().getData(Constants.LIBERTY_BUILD_FILE);
-            // if still null, prompt for user to select
-            if (buildFile == null) {
+            // if still null, and it is not from shift-shift, then prompt for user to select
+            if (isActionFromShiftShift || buildFile == null) {
```
mrglavas, line 74: In what context other than shift+shift would it be valid for a user to be given a dialog to choose from a list of projects? I think there was a case where there's only one project and this code just happens to work if no build file is available from the context. If there's more than one project, the user would be asked to select from a list, regardless of what they selected in the tool window.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -65,13 +65,16 @@ public void actionPerformed(@NotNull AnActionEvent e) {
...
                 List<LibertyModule> libertyModules = LibertyModules.getInstance()
                         .getLibertyModules(getSupportedProjectTypes());
-                if (!libertyModules.isEmpty()) {
+                if (isActionFromShiftShift || !libertyModules.isEmpty()) {
```
mrglavas, line 77: Consider shift+shift and no Liberty modules. A dialog asking the user to select from an empty list? An exception for passing an empty array to the list dialog? I haven't checked what happens because it wasn't possible to enter this block before if the list was empty.

fabio-franco: Good catch! The change in this if is not really necessary for the fix.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
fabio-franco: I think that the commented code I included is misleading... it should be something like `if still null, or it is from shift-shift, then prompt for the user to select`. This verification is covering two scenarios... where there is no build file in the context yet, and then it can be set, or it is already a build file provided but we can't rely on that because it could be just selected from the Liberty panel, but we are on a shift-shift action. Resuming, this code is shared between the shift-shift and the common right-click on "Start dev mode" in Liberty panel, so the verification would cover both cases.

kathrynkodama (approved):

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -67,11 +67,16 @@ public void actionPerformed(@NotNull AnActionEvent e) {
                 return;
             }
         }
-        if (buildFile == null) {
+
+        boolean isActionFromShiftShift = "GoToAction".equalsIgnoreCase(e.getPlace());
+
+        if (isActionFromShiftShift || buildFile == null) {
```
TrevCraw, line 73: If `isActionFromShiftShift` is true, do we need to assign a value to `buildFile` here, or is that only in the case where `buildFile` is null? It seems to me that we only care about assigning the `buildFile` at this point if `buildFile` was null to start (in other words, `isActionFromShiftShift` doesn't seem to have an affect on whether or not we should assign a value to the `buildFile`.)

If my assumption is correct, I would recommend taking out the `isActionFromShiftShift` boolean from the first `if` statement:
```
if (buildFile == null) {
    // set build file
}

if (isActionFromShiftShift || buildFile == null) {
    // prompt user for selection
}
 ```

fabio-franco: I just did the adjustment for it. Looking at GitHub changes can give the impression that it has more changes now but it has not. It is just that the if's are not nested anymore.

TrevCraw (approved):

Issue #129, opened 2022-11-09
Persist manually added or removed projects across IntelliJ sessions


TrevCraw:
Any project that has been manually added or removed should be persisted across IntelliJ work sessions. We need to store the list of manually added projects in IntelliJ metadata. And any time that a manually added (or other) project is removed, it should be removed from the IntelliJ metadata project list.

Dependency on #71 and #128

PR #170, opened 2022-11-24
Persist manually added/removed projects across IntelliJ sessions.

mrglavas:
Resolves: https://github.com/OpenLiberty/liberty-tools-intellij/issues/129

Example of `liberty-project-settings.xml`:
```
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="LibertyProjectSettings">
    <option name="customLibertyProjects">
      <set>
        <option value="$PROJECT_DIR$/start/pom.xml" />
      </set>
    </option>
  </component>
</project>
```

IntelliJ is parameterizing the path value when it saves it to the file. This should mean that the user can move their IntelliJ project to another directory after they've manually added/removed Liberty projects and it should still be able to resolve the persisted path.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/LibertyProjectSettings.java
+++ b/src/main/java/io/openliberty/tools/intellij/LibertyProjectSettings.java
@@ -1,0 +1,57 @@
+/*******************************************************************************
+ * Copyright (c) 2020, 2022 IBM Corporation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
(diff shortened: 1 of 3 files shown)


src/main/java/io/openliberty/tools/intellij/LibertyProjectSettings.java
```diff
@@ -0,0 +1,57 @@
+/*******************************************************************************
+ * Copyright (c) 2020, 2022 IBM Corporation.
```
kathrynkodama, line 2: Header should be just 2022

kathrynkodama (approved):
LGTM - just one minor copyright update needed

src/main/java/io/openliberty/tools/intellij/LibertyProjectSettings.java
mrglavas: @kathrynkodama Fixed.

fabio-franco (approved):

PR #174, opened 2022-11-25
Track terminal for each Liberty module

### LINKED ISSUE #143: Liberty Project terminal not working with newer versions of IntelliJ IDEA IDE
Related to the version bumps for supporting Java17 in PR #131.

Potentially two problems when using newer versions of IntelliJ IDE:
- Automatically opening up a dedicated terminal window(s) for the Liberty project(s) does not seem to work
- Running the dev commands via the Dashboard will open a new terminal, but cannot find the existing terminal when trying to stop.
image

After some extensive version testing, it seems this issue is only with IntelliJ 2022.2 and newer.
2022.1.4 is the last version where the terminal is working correctly with our current code. Luckily this still works with Java17 (but does not require it)

Not sure if there was a change in how the `ShellTerminalWidget` is made in the newer IntelliJ, or if it's just some regression in the IDE.

### Currently without Java17 support, on older IntelliJ versions
When we open a project, a terminal opens up with the project name. This is where the dev start/stop commands will run in. The commands will run in the dedicated project-name terminal without any issues.
image

### With IntelliJ 2022.2 and newer which require Java17
With newer versions of IntelliJ, somehow we are not able to open a dedicated terminal for the project. Running dev commands will open a new terminal, and `dev stop` will not be able to find the terminal where `dev start` was run. Present on both MacOS and Windows
image

kathrynkodama:
Fixes #143 

Associates each Liberty module with a ShellTerminalWidget (terminal instance). 

Also improved the check for whether we should run the stop dev mode and run tests action by using `widget.hasRunningCommands()`.  This returns true if there are running commands on the corresponding terminal instance. 

Signed-off-by: Kathryn Kodama <kathryn.s.kodama@gmail.com>

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
@@ -166,7 +166,6 @@ public class LibertyGeneralAction extends AnAction {
      */
     protected ShellTerminalWidget getTerminalWidget(boolean createWidget) {
         ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, libertyModule, createWidget);
-        libertyModule.setShellWidget(widget);
         if (widget == null || (!createWidget && !widget.hasRunningCommands())) {
             String msg;
             if (createWidget) {
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
@@ -166,6 +166,7 @@ public class LibertyGeneralAction extends AnAction {
      */
     protected ShellTerminalWidget getTerminalWidget(boolean createWidget) {
         ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, libertyModule, createWidget);
+        // Shows error for actions where terminal widget does not exist or action requires a terminal to already exist and expects "Start" to be running
         if (widget == null || (!createWidget && !widget.hasRunningCommands())) {
             String msg;
             if (createWidget) {
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
@@ -159,14 +159,15 @@ public class LibertyGeneralAction extends AnAction {
     }
 
     /**
-     * Returns the IntelliJ terminal widget for the corresponding Liberty module
+     * Returns the Terminal widget for the corresponding Liberty module
      *
-     * @param createWidget create terminal widget if it does not already exist
+     * @param createWidget create Terminal widget if it does not already exist
      * @return ShellTerminalWidget
      */
     protected ShellTerminalWidget getTerminalWidget(boolean createWidget) {
-        ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, projectName, createWidget);
-        if (widget == null) {
+        ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, libertyModule, createWidget);
+        libertyModule.setShellWidget(widget);
+        if (widget == null || (!createWidget && !widget.hasRunningCommands())) {
             String msg;
             if (createWidget) {
                 msg = LocalizedResourceUtil.getMessage("liberty.terminal.cannot.resolve", actionCmd, projectName);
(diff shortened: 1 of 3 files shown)


src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -159,14 +159,15 @@ protected void notifyError(String errMsg) {
...
      *
-     * @param createWidget create terminal widget if it does not already exist
+     * @param createWidget create Terminal widget if it does not already exist
      * @return ShellTerminalWidget
      */
     protected ShellTerminalWidget getTerminalWidget(boolean createWidget) {
-        ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, projectName, createWidget);
-        if (widget == null) {
+        ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, libertyModule, createWidget);
+        libertyModule.setShellWidget(widget);
```
evie-lau, line 169: I think `setShellWidget()` should be called at the time we create a new terminal in `LibertyProjectUtil.getTerminalWidget()` instead of every time we try to retrieve the ShellTerminalWidget.

evie-lau (changes requested):
Confirming that this works as intended, but I think a small optimization can be made per the comment above. I think we should only need to use `setShellWidget()` once when we create the terminal.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
kathrynkodama: Good idea, I've adjusted the code to incorporate this change

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
```diff
@@ -209,19 +201,28 @@ private static boolean isLibertyProject(PsiFile buildFile) {
...
+     *
+     * @param libertyModule
+     * @param terminalView
+     * @return ShellTerminalWidget or null if it does not exist
+     */
+    private static ShellTerminalWidget getTerminalWidget(LibertyModule libertyModule, TerminalView terminalView) {
+        ShellTerminalWidget widget = libertyModule.getShellWidget();
+        // check if widget exists in terminal view
+        if (widget != null) {
+            for (JBTerminalWidget terminalWidget : terminalView.getWidgets()) {
```
evie-lau, line 216: If someone were to close the associated terminal widget, and we cannot find it anymore, should we call `libertyModule.setShellWidget(null)`?

evie-lau: I don't think it'll affect the overall behavior, but rather just updates the state of the LibertyModule instance so we know it no longer has an open Shell Widget

kathrynkodama: I was thinking about this, which is why I was originally setting the widget each time we call `LibertyGeneralAction.getTerminalWidget()`.

However, we are already covered for this case by the method below, which will check if the terminal widget exists in the open terminal view. In the event that the user closes the terminal view themselves, this method will return null so it seems redundant to also set the libertyModule shell widget to null.
https://github.com/OpenLiberty/liberty-tools-intellij/blob/c438c4a2da73c6da5d9a5a32509666cb19aa70ac/src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java#L204-L223

kathrynkodama: For completeness, I adjusted the above method to call `libertyModule.setShellWidget(null);` if it returns null.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -159,14 +159,14 @@ protected void notifyError(String errMsg) {
...
      *
-     * @param createWidget create terminal widget if it does not already exist
+     * @param createWidget create Terminal widget if it does not already exist
      * @return ShellTerminalWidget
      */
     protected ShellTerminalWidget getTerminalWidget(boolean createWidget) {
-        ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, projectName, createWidget);
-        if (widget == null) {
+        ShellTerminalWidget widget = LibertyProjectUtil.getTerminalWidget(project, libertyModule, createWidget);
+        if (widget == null || (!createWidget && !widget.hasRunningCommands())) {
```
TrevCraw, line 169: suggestion to add a brief comment explaining the "createWidget" and "hasRunningCommands" logic

kathrynkodama: Addressed

TrevCraw (approved):

Issue #184, opened 2022-12-01
Exception: Cannot invoke "com.intellij.util.messages.MessageBusConnection.disconnect()" 


turkeylurkey:
Another red box in the lower right of the IDE. Caused when you switch from one project to another. 

When there is any exception starting the language server it will call `stop()` and this exception will most likely appear:
```
2022-12-01 10:52:34,273 [ 660399]   WARN - i.o.t.i.l.l.LanguageServiceAccessor - Cannot invoke "com.intellij.util.messages.MessageBusConnection.disconnect()" because "this.messageBusConnection" is null
java.lang.NullPointerException: Cannot invoke "com.intellij.util.messages.MessageBusConnection.disconnect()" because "this.messageBusConnection" is null
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.LanguageServerWrapper.stop(LanguageServerWrapper.java:431)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.LanguageServerWrapper.start(LanguageServerWrapper.java:347)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.LanguageServerWrapper.getInitializedServer(LanguageServerWrapper.java:659)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.LanguageServiceAccessor.lambda$getLanguageServers$15(LanguageServiceAccessor.java:577)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1845)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:575)
	at java.base/java.util.stream.AbstractPipeline.evaluateToArrayNode(AbstractPipeline.java:260)
	at java.base/java.util.stream.ReferencePipeline.toArray(ReferencePipeline.java:616)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.LanguageServiceAccessor.getLanguageServers(LanguageServiceAccessor.java:590)
[...]

PR #185, opened 2022-12-01
Check for null before using messageBusConnection.

turkeylurkey:
Exception on line 342 will cause messageBusConnection to be used before being set.

image

Signed-off-by: Paul Gooderham <turkeyonmarblerye@gmail.com>


turkeylurkey:
Fixes #184

evie-lau (approved):

Issue #197, opened 2022-12-06
NPE when trying to start on a Gradle project


kathrynkodama:
On Linux, using a gradle app generated by the open liberty starter
```
java.lang.NullPointerException: Cannot invoke "org.jetbrains.plugins.gradle.settings.GradleProjectSettings.getDistributionType()" because "gradleProjectSettings" is null
	at io.openliberty.tools.intellij.util.LibertyGradleUtil.getGradleSettingsCmd(LibertyGradleUtil.java:164)
	at io.openliberty.tools.intellij.actions.LibertyDevStartAction.executeLibertyAction(LibertyDevStartAction.java:36)
	at io.openliberty.tools.intellij.actions.LibertyGeneralAction.actionPerformed(LibertyGeneralAction.java:120)
	at io.openliberty.tools.intellij.LibertyExplorer.executeAction(LibertyExplorer.java:343)
	at io.openliberty.tools.intellij.LibertyExplorer$2.onDoubleClick(LibertyExplorer.java:258)
	at com.intellij.ui.DoubleClickListener.onClick(DoubleClickListener.java:30)
	at com.intellij.ui.ClickListener$1.mouseReleased(ClickListener.java:59)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:298)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6648)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3392)
	at com.intellij.ui.treeStructure.Tree.processMouseEvent(Tree.java:422)
	at java.desktop/java.awt.Component.processEvent(Component.java:6413)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5022)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4854)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4948)
	at java.desktop/java.awt.LightweightDispatcher.processMouseEvent(Container.java:4575)
	at java.desktop/java.awt.LightweightDispatcher.dispatchEvent(Container.java:4516)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2310)
[...]

PR #208, opened 2022-12-07
Including an exception for when gradle settings is null

fabio-franco:
This is a fix for https://github.com/OpenLiberty/liberty-tools-intellij/issues/197 following the strategy we agreed during scrum, that is, throw a nice message error in the case gradle settings is null.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
@@ -169,7 +169,7 @@ public class LibertyGradleUtil {
         GradleProjectSettings gradleProjectSettings = GradleSettings.getInstance(project).getLinkedProjectSettings(project.getBasePath());
         if (gradleProjectSettings == null) {
             String translatedMessage = LocalizedResourceUtil.getMessage("gradle.settings.is.null");
-            throw new LibertyException("Cannot execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.", translatedMessage);
+            throw new LibertyException("Could not execute action because there is an error with Gradle configuration. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.", translatedMessage);
         }
         else if (gradleProjectSettings.getDistributionType().isWrapped()) {
             // a wrapper will be used
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
@@ -167,7 +167,11 @@ public class LibertyGradleUtil {
      */
     public static String getGradleSettingsCmd(Project project) throws LibertyException {
         GradleProjectSettings gradleProjectSettings = GradleSettings.getInstance(project).getLinkedProjectSettings(project.getBasePath());
-        if (gradleProjectSettings.getDistributionType().isWrapped()) {
+        if (gradleProjectSettings == null) {
+            String translatedMessage = LocalizedResourceUtil.getMessage("gradle.settings.is.null");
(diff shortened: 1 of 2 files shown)


fabio-franco:
Hi @mrglavas and @dshimo, when you have a chance, can you verify if this pr covers the issue mentioned on scrum today. So as agreed, it should show a pop up message instead of the null pointer exception.

fabio-franco:
@TrevCraw, I am wondering if I could change the message from `Cannot execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.` to `Could not execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.` to reflect the style used on the other maven/gradle messages. What do you think?

src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
```diff
@@ -167,7 +167,11 @@ private static boolean containerVersion(String plugin) {
      */
     public static String getGradleSettingsCmd(Project project) throws LibertyException {
         GradleProjectSettings gradleProjectSettings = GradleSettings.getInstance(project).getLinkedProjectSettings(project.getBasePath());
-        if (gradleProjectSettings.getDistributionType().isWrapped()) {
+        if (gradleProjectSettings == null) {
+            String translatedMessage = LocalizedResourceUtil.getMessage("gradle.settings.is.null");
+            throw new LibertyException("Cannot execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.", translatedMessage);
```
TrevCraw, line 172:
```suggestion
            throw new LibertyException("Could not execute action because there is an error with Gradle configuration. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.", translatedMessage);
```

src/main/resources/messages/LibertyBundles.properties
```diff
@@ -134,4 +134,5 @@ gradle.wrapper.does.not.exist=A Gradle wrapper for the project could not be foun
 gradle.wrapper.cannot.execute=Could not execute Gradle wrapper because the process does not have permission to execute it. Consider giving executable permission for the Gradle wrapper file or changing the build preferences for Gradle inside IntelliJ Gradle preferences.
 gradle.invalid.build.preference=Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.
 gradle.cannot.execute=Could not execute Gradle from {0} because the process does not have permission to execute it. Consider giving executable permission for the Gradle executable or configure IntelliJ to use the Gradle wrapper.
-gradle.does.not.exist=Could not execute the Gradle executable {0}. Make sure a valid path is configured inside IntelliJ Gradle preferences.
\ No newline at end of file
+gradle.does.not.exist=Could not execute the Gradle executable {0}. Make sure a valid path is configured inside IntelliJ Gradle preferences.
+gradle.settings.is.null=Cannot execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.
```
TrevCraw, line 138:
```suggestion
gradle.settings.is.null=Could not execute action because there is an error with Gradle configuration. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.
```

TrevCraw:
> @TrevCraw, I am wondering if I could change the message from `Cannot execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.` to `Could not execute project because there is an error with configured Gradle. Make sure to configure a valid path for Gradle inside IntelliJ Gradle preferences.` to reflect the style used on the other maven/gradle messages. What do you think?

@fabio-franco Good idea. I included that change as part of my suggestions.

mrglavas (approved):

TrevCraw (approved):

PR #213, opened 2022-12-08
Fix launching LS with IntelliJ 2022.3

evie-lau:
Fix another warning/exception with ActionPlaces

Tested and runs properly on 2022.3 in MacOS and Windows

### DIFF
--- a/src/main/resources/META-INF/plugin.xml
+++ b/src/main/resources/META-INF/plugin.xml
@@ -29,7 +29,7 @@
     <depends>com.intellij.properties</depends>
     <depends>org.jetbrains.idea.maven</depends>
     <depends>com.intellij.gradle</depends>
-    <idea-version since-build="222" until-build="223"/>
+    <idea-version since-build="222" until-build="223.*"/>
 
     <extensions defaultExtensionNs="com.intellij">
         <toolWindow anchor="right" id="Liberty" icon="/icons/OL_logo_13.svg"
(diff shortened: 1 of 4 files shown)


kathrynkodama (approved):
LGTM, thanks @evie-lau

src/main/resources/META-INF/plugin.xml
```diff
@@ -29,7 +29,7 @@
     <depends>com.intellij.properties</depends>
     <depends>org.jetbrains.idea.maven</depends>
     <depends>com.intellij.gradle</depends>
-    <idea-version since-build="222" until-build="222.*"/>
+    <idea-version since-build="222"/>
```
kathrynkodama, line 32: Should we preemptively block for anything past 223.* or wait to see if it breaks us?

evie-lau: I was considering that as well. Given how unstable LSP integration with IntelliJ is, might be better to add it in and bump when available and working.

src/main/resources/META-INF/plugin.xml
```diff
@@ -29,7 +29,7 @@
     <depends>com.intellij.properties</depends>
     <depends>org.jetbrains.idea.maven</depends>
     <depends>com.intellij.gradle</depends>
-    <idea-version since-build="222" until-build="222.*"/>
+    <idea-version since-build="222" until-build="223"/>
```
kathrynkodama, line 32: Do we have to use `223.*` to include all versions of 223? Looking at https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html seems like yes?

evie-lau: 👍🏼 Seems like a safe bet.

kathrynkodama (approved):

Issue #147, opened 2022-11-22
Hover feature not working for LCLS & LemMinX, MicroProfile properties files


TrevCraw:
The hover feature is not working on any platform for LCLS and LemMinX.

It is suspected that this exception, which occurs over and over, could be the underlying problem:

```
java.lang.IllegalArgumentException: Invalid range specified: (31, 26); 
	at com.intellij.openapi.util.TextRange.assertProperRange(TextRange.java:231)
	at com.intellij.openapi.util.TextRange.assertProperRange(TextRange.java:226)
	at com.intellij.openapi.util.TextRange.assertProperRange(TextRange.java:222)
	at com.intellij.openapi.util.TextRange.<init>(TextRange.java:41)
	at com.intellij.openapi.util.TextRange.<init>(TextRange.java:30)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.operations.completion.LSIncompleteCompletionProposal.getPrefixCompletionStart(LSIncompleteCompletionProposal.java:163)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.operations.completion.LSIncompleteCompletionProposal.<init>(LSIncompleteCompletionProposal.java:84)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.operations.completion.LSCompletionProposal.<init>(LSCompletionProposal.java:26)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.operations.completion.LSContentAssistProcessor.createLookupItem(LSContentAssistProcessor.java:104)
	at io.openliberty.tools.intellij.lsp4mp.lsp4ij.operations.completion.LSContentAssistProcessor.lambda$toProposals$4(LSContentAssistProcessor.java:92)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:197)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1625)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #244, opened 2023-01-16
Start language servers based on file patterns

kathrynkodama:
Part of #147 

Enables hover for bootstrap.properties and microprofile-config.properties files while keeping the correct completion items for each file type.

Modifies `languageMapping` extension point to also accept a file pattern, and filters on that file pattern when determining which language servers should activate/be active on each file.

Also added some logging messages for language server logic, LOGGER.info messages show up in the idea.log.


src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/LanguageServerWrapper.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2022 Red Hat, Inc.
+ * Copyright (c) 2020, 2023 Red Hat, Inc.
```
turkeylurkey, line 2: I think you should leave the Red Hat copyright alone and add a new line saying "(C) Copyright IBM Corp. 2023" or put Copyright first to match Red Hat's.

turkeylurkey: Oops I actually meant this for LanguageServersRegistry.java

turkeylurkey (approved):

src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/LanguageServerWrapper.java
kathrynkodama: Resolving this for now, to re-visit once we have more guidance

turkeylurkey (approved):

Issue #228, opened 2022-12-14
Empty dropdown when trying to add a Gradle project using shift-shift on Windows


fabio-franco:
For this test, I used liberty-bikes project. This issue may be related to this issue https://github.com/OpenLiberty/liberty-tools-intellij/issues/182, that is, the fact that liberty-bikes inside IntelliJ is showing all the project with the same label:

<img width="145" alt="image" src="https://user-images.githubusercontent.com/7606936/207698993-7067f157-6182-486d-a695-c038bacf9d53.png">

To reproduce the issue, I renamed the "config" folder in the `player-service` and then it correctly disappeared from the menu. However, when clicking on shift-shift to add the project again, after I chose "Add Project to selected window", and then it appeared a dropdown with an empty select option:

<img width="273" alt="image" src="https://user-images.githubusercontent.com/7606936/207681216-ea02b8c4-a627-4521-a491-5bea5cb68935.png">

<img width="1013" alt="image" src="https://user-images.githubusercontent.com/7606936/207680204-33f32420-c086-4414-82fa-e2596b636521.png">

TrevCraw:
Attempt to reproduce with multiple single module projects under same folder.

TrevCraw:
Closed as part of https://github.com/OpenLiberty/liberty-tools-intellij/pull/263

PR #263, opened 2023-02-08
Update Gradle project name

### LINKED ISSUE #182: Project with multi modules is showing all modules with same name [Gradle]
When using the liberty-bikes project (https://github.com/OpenLiberty/liberty-bikes), that is a multi-module project, Intellij is showing all the modules with the same name, which can confuse the end-user. See the image below:

image

A suggestion, in case of gradle, is to use the name inside the `build.grade` file.

TrevCraw:
Fixes #182 and #228 
Use name of directory containing the build file if settings.gradle is not found (Gradle) or if artifactId is not found (Maven)

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
+++ b/src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
@@ -120,8 +120,13 @@ public class LibertyExplorer extends SimpleToolWindowPanel {
                 LOGGER.warn(String.format("Could not resolve project name from build file: %s", virtualFile), e);
             }
             if (projectName == null) {
-                projectName = project.getName();
+                if (virtualFile.getParent() != null) {
+                    projectName = virtualFile.getParent().getName();
+                } else {
+                    projectName = project.getName();
+                }
             }
+
             boolean validContainerVersion = buildFile.isValidContainerVersion();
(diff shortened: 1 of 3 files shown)


kathrynkodama:
@TrevCraw Will we still run into #228 when two projects have the same name (either in settings.gradle or based on the build file dir name)?

TrevCraw:
> @TrevCraw Will we still run into #228 when two projects have the same name (either in settings.gradle or based on the build file dir name)?

I will test that out. My assumption is that it would just be two dropdown selections with the same name.

TrevCraw:
This is how things look with the code changes using two Gradle projects in the same workspace with either the same project name in `settings.gradle` or the same dir name containing `build.gradle`.

With the same project name in `settings.gradle`

- One project undetected by the Liberty Tool window
screenshot-1
- Both projects undetected by the Liberty Tool window
screenshot-2


With no `settings.gradle` in either project and the same dir name containing the `build.gradle`

- One project undetected by the Liberty Tool window
screenshot-3
- Both projects undetected by the Liberty Tool window
screenshot-4

This seems like the correct behaviour. The fact that the projects are indistinguishable by name comes down to how the user has configured their projects. I'm not sure there is anything more that Liberty Tools should do for this scenario.

TrevCraw:
Regarding #182 , this is how the [liberty-bikes](https://github.com/OpenLiberty/liberty-bikes) project now looks:
liberty-bikes-1

And if you make a project undetectable by the Liberty Tool window, this is how it looks from the "shift-shift" window trying to add projects to the dashboard:
liberty-bikes-2

And if you make a second project undetectable:
liberty-bikes-3

kathrynkodama (approved):

PR #265, opened 2023-02-10
Fix for diagnostics not appearing until another LS capability is triggered

### LINKED ISSUE #258: Diagnostics aren't appearing until another LS capability is triggered
On opening a file, the diagnostics are not populated until you first hover or trigger completion

kathrynkodama:
Fixes #258 
- Call `getLanguageServers()` instead of `getLSWrappers()` in `LSPLocalInspectionTool.checkFile()` (inspection/diagnostic logic) to ensure language servers are properly initialized and the correct textDocument/didOpen requests are sent to language servers. 
- Updates org.jetbrains.intellij plugin to 1.12.0
- Add commented out logic to attach JVM debugger to LCLS

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id 'java'
-    id 'org.jetbrains.intellij' version '1.11.0'
+    id 'org.jetbrains.intellij' version '1.12.0'
 }
 
 group 'io.openliberty.tools'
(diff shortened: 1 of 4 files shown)


TrevCraw (approved):

turkeylurkey (approved):
Never noticed this before but the method previously called was described as "This method is currently internal and should only be referenced for testing." Good find.

Issue #222, opened 2022-12-13
Add hover info to entries in the Liberty Tool window


turkeylurkey:
There could be multiple apps in the window and it should be easy to tell which entry corresponds to which project.

![image](https://user-images.githubusercontent.com/24705144/207199506-9ce6ca6d-2363-436f-a620-b483f3bf5f08.png)

Perhaps if you hover over the element it should display the full path of the build file (pom.xml, build.gradle)

Rahul-Biju:
I have used a MouseMotionListener which is an interface in Java Swing that provides methods to handle mouse motion events, such as moving the mouse.So now what happens is If the mouse is not over the project then the path won’t show. If the mouse is over the project then the path will show.

issue : at some points in the liberty dashboard , when the cursor moves the path is showing.  

Next step : this issue need to be addressed.

PR #269, opened 2023-02-21
Issue 137

ajm01:
in the case where there are multiple applications in the dashboard with the same name (think the start and finish apps in the guides) the Shift-Shift choice of liberty dashboard actions presents a dialog chooser that contains identical entries. This fix incorporates a tooltip hover action that displays the path to the build file for each app listed so the user may determine which one to choose. 

in the course of adding this i determined that the chooser dialog was always returning the initalValue set in the list when the same named app appeared more than once. this is because the underlying Dialog/Chooser implementation is matching the LibertyModule name it is looking for with the first name it finds in the list of choices when determining the index of the correct choice. i had to add an additional piece of data - the folder name that contains the build file from the buildPath string. This allows the Dialog matcher to match with the distinct app that the user chose from the list.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyProjectChooserDialog.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyProjectChooserDialog.java
@@ -1,3 +1,12 @@
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
 package io.openliberty.tools.intellij.actions;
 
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/intellij/actions/LibertyProjectChooserDialog.java
```diff
@@ -0,0 +1,88 @@
+package io.openliberty.tools.intellij.actions;
```
kathrynkodama, line 1: Missing license header

kathrynkodama (commented):
@ajm01 we should consider adding the same tool tip enhancement to the entries in the Liberty tool window - hover for the full build file path (could be a new enhancement issue if it's too much to include in this fix).

A bit of a design choice - should we use the new project name format ` <app-name> : <buildfile parent folder>` in the Liberty tool window as well? I can open a separate issue for this if we would like - that would be applicable across the 3 IDEs for consistency.

TrevCraw:
Enhancement issue for adding tool tip with full build file path: https://github.com/OpenLiberty/liberty-tools-intellij/issues/222

Issue #85, opened 2022-09-01
LS Integration: Document syncing on diagnostics


kathrynkodama:
Document seems to fall out of sync, outdated diagnostics are still being shown after they are fixed.

For example, valid diagnostic:

<img width="902" alt="image" src="https://user-images.githubusercontent.com/26146482/187963545-6d7dfb47-8741-45f5-ae32-2329da2ed0a5.png">

After adding the corresponding method, the outdated diagnostic is still being shown:
<img width="985" alt="Screen Shot 2022-09-01 at 12 19 00 PM" src="https://user-images.githubusercontent.com/26146482/187963724-da056fbe-ddce-400d-8692-66b2d2bbc68b.png">

turkeylurkey:
This is a clarification of the previous comment. When you first open the editor there is no diagnostic:
<img width="680" alt="image" src="https://user-images.githubusercontent.com/24705144/199616327-4a9937c3-8ba4-4c6d-9e6a-5f15cc5ef8aa.png">
Then hit one character and the diagnostic appears as expected, there is no one keystroke difference:
<img width="745" alt="image" src="https://user-images.githubusercontent.com/24705144/199617569-57102fd4-f5df-45e2-aa53-bf648755567b.png">
However there is no hover information at this point. One keystroke (backspace) later the diagnostic is rendered differently (all red):
<img width="938" alt="image" src="https://user-images.githubusercontent.com/24705144/199617770-0e2ee8b6-4233-477e-9187-2e5479ebe651.png">
And the hover information is one keystroke behind.

The diagnostic is generated when the language server receives the message microprofile/java/diagnostics as specified in MicroProfileJavaDiagnosticsProvider implemented in MicroProfileLanguageClient:
```
@Override
  public CompletableFuture<List<PublishDiagnosticsParams>> getJavaDiagnostics(MicroProfileJavaDiagnosticsParams javaParams) {
    return runAsBackground("Computing Java diagnostics", () -> PropertiesManagerForJava.getInstance().diagnostics(javaParams, PsiUtilsLSImpl.getInstance(getProject())));
  }
```
PropertiesManagerForJava collects the class name from an extension point:
```
List<IJavaDiagnosticsParticipant> definitions = IJavaDiagnosticsParticipant.EP_NAME.extensions()
```
see 
```
[...]

PR #277, opened 2023-02-28
Refresh inspections to clear old diagnostics

kathrynkodama:
Final fix for #85


dshimo (approved):

evie-lau (approved):
It does seem to fix it by fixing the original intent of the forced `checkFile` run to run after everything. (Order is accept(publishDiagnostics) -> createMarkers -> force checkFile)

But it seems the initial issue for the doc syncing was happening due to a race condition of `checkFile` getting run first, before `accept(publishDiagnostics)`.

PR #279, opened 2023-03-02
Fixes #275 for the most part.

### LINKED ISSUE #275: LS Integration: XML comment diagnostics forces document into a broken state
**To reproduce**
1. Open to a server.xml file
2. Comment out a line - I used the keyboard shortcut <kbd>Cmd</kbd> + <kbd>/</kbd>.Diagnostic appears on the next comment in the document indicating that "The string "--" is not permitted within comments." In the screenshot below, I used the keyboard shortcut to comment out line 15, the diagnostic appeared on the comment on line 17.
image
3. Enter into a broken state, switching windows does not correct the issue. 

Able to reproduce on Mac, Linux and Windows.

evie-lau:
Fix the missing/overwritten `changeParams` when there are multiple change events created simultaneously.

This seems to work most of the time, but there are some edge cases where the language server processes the `didChange` event out of order, despite this fix in getting them sent properly and in order in `DocumentContentSynchronizer.java` - seems to happen mostly soon after starting the language server.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/DocumentContentSynchronizer.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/DocumentContentSynchronizer.java
@@ -26,7 +26,9 @@ import org.slf4j.LoggerFactory;
 
 import javax.annotation.Nonnull;
 import java.net.URI;
+import java.util.ArrayDeque;
 import java.util.Collections;
+import java.util.Deque;
 import java.util.concurrent.CompletableFuture;
 
 public class DocumentContentSynchronizer implements DocumentListener {
@@ -38,7 +40,7 @@ public class DocumentContentSynchronizer implements DocumentListener {
     private final TextDocumentSyncKind syncKind;
 
     private int version = 0;
-    private DidChangeTextDocumentParams changeParams;
+    private Deque<DidChangeTextDocumentParams> changeParamsQueue = new ArrayDeque<>();
     private long modificationStamp;
     final @Nonnull CompletableFuture<Void> didOpenFuture;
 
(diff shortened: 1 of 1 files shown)


kathrynkodama (approved):
LGTM and a huge improvement from where it was!

angelozerr:
@evie-lau I discovered that you have copied/pasted the code from https://github.com/redhat-developer/intellij-quarkus and you did some fix in your repository but I think we should work together.

For instance I did a similar fix for this issue with the PR https://github.com/redhat-developer/intellij-quarkus/pull/778/files but I send several change events compare to your PR. I spend a lot of time to play with this PR with our Qute support (which looks like our XML language server) and I have never seen some trouble. Please test it if you think it is better, please apply it to have the same code base than us.

IMHO I think we should have a commons robust IJ LSP plugin and to provide that we should have one base code reference. As we have similar bug, I suggest that you create the issue in https://github.com/redhat-developer/intellij-quarkus too and reference your issue. I think the fix should be done in https://github.com/redhat-developer/intellij-quarkus to have a code reference for LSP (and LSP4MP too, we will align LSP4MP to last version soon).

evie-lau:
@angelozerr your fix did indeed work better and never gets out of sync anymore. Will be updating this code to use your fix.

And agreed on having a common IJ LSP repo/plugin. Our team has had discussions in the past with @kathrynkodama regarding separating that, although she is no longer actively working on these projects now.

angelozerr:
> your fix did indeed work better and never gets out of sync anymore. Will be updating this code to use your fix.

Glad my fix works good for you.

> And agreed on having a common IJ LSP repo/plugin. Our team has had discussions in the past with @kathrynkodama regarding separating that, although she is no longer actively working on these projects now.

Glad you like this idea. We have not the time for the moment to provide this Kind of plugin but my plan is to fix all problems with LSP on IJ, improving some LSP opération like snippet support for completion, implement workspace symbols etc.

My current work is to improve diagnostics and code action to improve à lot the performance and fix buts with outsync diagnostics with robust mean.

So please trace on our IJ plugin to have the same base code and dont hesitate to review my PRs and create some PAS with fix to have the sale base code. Thansk!

PR #292, opened 2023-03-14
hover tests in server.xml

ajm01:


src/test/java/io/openliberty/tools/intellij/it/GradleSingleModLSTest.java
```diff
@@ -0,0 +1,38 @@
+package io.openliberty.tools.intellij.it;
```
mezarin, line 1: Not necessarily part of your changes, but could you please add the test-automation branch to the list of branches here .github/workflows/build.yaml:
```
on:
  push:
    branches: '**'
  pull_request:
    branches: [main, ls-integration]
```
That will allow your tests to run on Linux and Windows.

PR #303, opened 2023-03-27
Add ManagedBeanConstructorQuickFix.

turkeylurkey:
This quick fix enables the addition of the @Inject annotation on a constructor as demonstrated in the test case: jakarta.jdt/org.eclipse.lsp4jakarta.jdt.test/projects/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/cdi/ManagedBeanConstructor.java

This code assumes IntelliJ will add the import statement as part of reformatting. I believe this dependency is also part of the MicroProfile quick fixes. 

Also assuming we do not update the copyright if there were no changes other than the package name. Feel free to comment.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanConstructorQuickFix.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanConstructorQuickFix.java
@@ -1,0 +1,20 @@
+/* Copyright (c) 2021 IBM Corporation.
+*
+* This program and the accompanying materials are made available under the
+* terms of the Eclipse Public License v. 2.0 which is available at
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Hani Damlaj
+*******************************************************************************/
+package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.cdi;
+
+import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction.proposal.quickfix.InsertAnnotationMissingQuickFix;
+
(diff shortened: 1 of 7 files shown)


turkeylurkey:
Part of #304

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanConstructorQuickFix.java
```diff
@@ -0,0 +1,20 @@
...
+* http://www.eclipse.org/legal/epl-2.0.
+*
+* SPDX-License-Identifier: EPL-2.0
+*
+* Contributors:
+*     Hani Damlaj
+*******************************************************************************/
+package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.cdi;
+
+import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction.proposal.quickfix.InsertAnnotationMissingQuickFix;
```
evie-lau, line 14: Was there any problem with using the existing `io.openliberty.tools.intellij.lsp4mp4ij.psi.core.java.codeaction.InsertAnnotationMissingQuickFix` ?
I noticed the classes are slightly different.

turkeylurkey: To use the lsp4mp quick fixes we would have to adopt their architecture. For instance they put some code in getCodeActions() and the rest of it in resolveCodeAction(). There is other code that decides whether to call the language server to resolve a code action or to call resolveCodeAction() without that. My thinking is not to change the architecture of the lsp4jakarta quick fixes for now. If you think the simpler implementation used by lsp4jakarta wont work with IntelliJ let's have a discussion.

evie-lau (approved):

dshimo (approved):

turkeylurkey:
AddPathParamQuickFix fixes #304

PR #307, opened 2023-03-28
Fixes #288 - fix ExternalAnnotator sync issues, disable LocalInspectionTool

### LINKED ISSUE #288: Document sync issue from new diagnostic style
The merge with the LSP4MP 0.6.0 release caused diagnostic sync issues (from the diagnostics provided via <externalAnnotator> which uses the new LSPCodeActionAnnotator class, which has similar logic to the checkFile method currently have to force rerun to fix syncing)

The diagnostic that comes from LSPLocalInspectionTool is still fine from our fix. only the diagnostic from the externalAnnotator has sync issues

The symptom is that when you first fix the indicated code error the diagnostic is not removed but if you change the file again in any way the diagnostic is removed. Similarly, instead of changing the file a second time you can select a different application on the PC and when you return to IntelliJ the diagnostic will be removed.

evie-lau:
Move all logic to run in the `apply` step, which seems to work in our favor for the timing of when it runs compared to our diagnostics getting accepted. Seems to fix the syncing issues.


dshimo (approved):
moving into apply logic :+1:
can hopefully remove checkFile

Issue #122, opened 2022-11-02
Consume quick fixes for LSP4MP


TrevCraw:
Find all the quick fix Java files and make list in this issue. Rank in order of difficulty and sort by similarity. Start with the easier ones first.
Consider quick fixes for both Java and properties files.
dependency on: https://github.com/OpenLiberty/liberty-tools-intellij/issues/81
Need: org.eclipse.lsp4mp.jdt.internal.core.java.codeaction.CodeActionHandler

List of Eclipse LSP4MP quick-fix classes (to be Ranked):

Fixed in #320
- [x] ApplicationScopedAnnotationMissingQuickFix
- [x] HealthAnnotationMissingQuickFix
- [x] InsertAnnotationMissingQuickFix
- [x] MicroProfileGenerateOpenAPIOperation
- [x] NoValueAssignedToPropertyQuickFix - need to make sure project has a `META-INF/microprofile-config.properties` file. the MP starter project did not have this file in the META-INF directory

Fixed in #286 
- [x] ImplementHealthCheckQuickFix
- [x] InsertAnnotationAttributeQuickFix

dshimo:
The `plugin.xml` helps pair a diagnostic to its associated quick fix.

```
    <javaCodeActionParticipant kind="quickfix"
            targetDiagnostic="microprofile-config#NO_VALUE_ASSIGNED_TO_PROPERTY"
            implementationClass="io.openliberty.tools.intellij.lsp4mp4ij.psi.internal.config.java.NoValueAssignedToPropertyQuickFix"/>
    <javaCodeActionParticipant kind="quickfix"
            targetDiagnostic="microprofile-config#NO_VALUE_ASSIGNED_TO_PROPERTY"
            implementationClass="io.openliberty.tools.intellij.lsp4mp4ij.psi.internal.config.java.InsertDefaultValueAnnotationAttributeQuickFix"/>
    <javaCodeActionParticipant kind="quickfix"
            targetDiagnostic="microprofile-health#ImplementHealthCheck"
            implementationClass="io.openliberty.tools.intellij.lsp4mp4ij.psi.internal.health.java.ImplementHealthCheckQuickFix"/>
    <javaCodeActionParticipant kind="quickfix"
            targetDiagnostic="microprofile-health#HealthAnnotationMissing"
            implementationClass="io.openliberty.tools.intellij.lsp4mp4ij.psi.internal.health.java.HealthAnnotationMissingQuickFix"/>
[...]

PR #320, opened 2023-03-30
Try/catch out of bounds when applying text edits.

evie-lau:
The `toOffset` method would get out of bounds because the it was using the line numbers from the new text, but trying to calculate using the pre-change document.

Added a try/catch to use the end offset of the original document instead.

Fixes some of the quickfixes in #122 
- HealthAnnotationMissingQuickFix
- InsertAnnotationMissingQuickFix
- MicroProfileGenerateOpenAPIOperation
- ApplicationScopedAnnotationMissingQuickFix

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/LSPIJUtils.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/LSPIJUtils.java
@@ -215,12 +215,18 @@ public class LSPIJUtils {
         for(TextEdit edit : edits) {
             if (edit.getRange() != null) {
                 String text = edit.getNewText();
+                // compute start and end char offsets of the new Edit text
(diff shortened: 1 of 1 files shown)


evie-lau:
Looking into a seemingly resurfaced issue. Tried manually editing/replacing some text in a `server.xml`, and seem to be running into some doc sync issues that was resolved via #279 introducing the queue of doc changes.

Potentially a timing/race condition for the doc changes being applied to the editor, and the `accept` diagnostics code getting the latest copy of the document.

Some example scenarios that happened:
- had location="test1.xml", highlighted location, typed a -> read n is not allowed in element 'include'
- had b="text1.xml" , highlighted b, typed location -> read bocation is not allowed...
also seems to be inconsistent, and sometimes fixes itself with more document edits.

dshimo (approved):

Issue #310, opened 2023-03-28
Quick fixes related to RemoveAnnotationConflictQuickFix


turkeylurkey:
After enabling RemoveAnnotationConflictQuickFix also implement:
- [x] RemoveAnnotationConflictQuickFix
- [x] RemoveMultipleAnnotations
- [x] JsonbTransientAnnotationQuickFix
- [x] ScopeDeclarationQuickFix

Other quick fixes which extend RemoveAnnotationConflictQuickFix. Enable and test:
- [x] ConflictInjectMultipleConstructorQuickFix?
- [x] ConflictProducesInjectQuickFix
- [x] DeleteConflictMapKeyQuickFix
- [x] JsonbAnnotationQuickFix
- [x] RemoveInjectAnnotationQuickFix
- [x] RemovePostConstructAnnotationQuickFix
- [x] RemovePreDestroyAnnotationQuickFix
- [x] RemoveProduceAnnotationQuickFix

PR #337, opened 2023-04-12
Quick fixes related to RemoveAnnotationConflictQuickFix

turkeylurkey:
I had to decide between using the new RemoveAnnotationsProposal and porting over the existing DeleteAnnotationProposal and I decided that the loop to find the annotation model objects to pass to the former was just about as difficult as porting the latter. Let me know what you think. 

Not finished with #310 yet.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/proposal/DeleteAnnotationProposal.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/proposal/DeleteAnnotationProposal.java
@@ -14,10 +14,7 @@
 package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction.proposal;
 
 import com.intellij.openapi.editor.Document;
-import com.intellij.psi.PsiAnnotation;
-import com.intellij.psi.PsiElement;
-import com.intellij.psi.PsiFile;
-import com.intellij.psi.PsiVariable;
+import com.intellij.psi.*;
 import io.openliberty.tools.intellij.lsp4mp4ij.psi.core.java.corrections.proposal.Change;
 import io.openliberty.tools.intellij.lsp4mp4ij.psi.core.java.corrections.proposal.ChangeCorrectionProposal;
 import org.eclipse.lsp4j.CodeActionKind;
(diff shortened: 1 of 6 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/proposal/quickfix/RemoveAnnotationConflictQuickFix.java
```diff
@@ -0,0 +1,132 @@
...
+
+    }
+
+    protected void removeAnnotations(Diagnostic diagnostic, JavaCodeActionContext context, PsiElement parentType,
+                                     List<CodeAction> codeActions) {
+        if (generateOnlyOneCodeAction) {
+            removeAnnotation(diagnostic, context, parentType, codeActions, annotations);
+        } else {
+            for (String annotation : annotations) {
+                removeAnnotation(diagnostic, context, parentType, codeActions, annotation);
```
turkeylurkey, line 87: Before anybody mentions I just noticed that in this loop we need to create a new context so that will come in a future pull request.

turkeylurkey: Never mind, I can add it to this pull request.

dshimo (approved):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/proposal/DeleteAnnotationProposal.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public Change getChange() {
+        if (declaringNode instanceof PsiModifierListOwner) {
+            PsiModifierListOwner targetNode = ((PsiModifierListOwner) declaringNode);
+            PsiAnnotation[] targetAnnotations = targetNode.getAnnotations();
+            for (var annotation : targetAnnotations) {
+                // Allow the names in targetAnnotations to be fully qualified or short (no package name).
+                if (Arrays.stream(annotations).anyMatch(a -> annotation.getQualifiedName().endsWith(a))) {
```
mrglavas, line 67: I'm guessing the quick fixes from Eclipse behaved like this, but in general this seems like the wrong thing to be doing. If this code is only meant to be deleting Jakarta EE related annotations, then if "a" equals "QueryParam", that could delete "@my.unrelated.QueryParam" when you only intended to remove "@jakarta.ws.rs.QueryParam". They're completely different annotations.

mrglavas: Also imagine if there were two distinct annotations, "Param" and "QueryParam", if "a" equals "Param" that would delete both annotations, where you probably only meant to delete "Param".

turkeylurkey: Ok, added getFQAnnotationNames() to look up the annotations by short name. Take a look.

mrglavas: I think the thing I'm missing is how there are any short names to resolve. As I've been porting quick fixes, I've been replacing short names I find with their fully qualified names to avoid issues with matching ambiguous names. I think there's code that I don't see yet in a follow-up PR you're still preparing. Going to approve this one so you can move forward.

PR #339, opened 2023-04-14
Add quick fixes which simply extend RemoveAnnotationConflictQuickFix.

turkeylurkey:
Mostly one liners that leverage previous work.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/JakartaCodeActionHandler.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/JakartaCodeActionHandler.java
@@ -14,16 +14,21 @@ package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction;
 import com.intellij.psi.PsiFile;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.annotations.AnnotationConstants;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.annotations.PostConstructReturnTypeQuickFix;
+import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.annotations.RemovePostConstructAnnotationQuickFix;
+import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.annotations.RemovePreDestroyAnnotationQuickFix;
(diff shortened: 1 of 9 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/JakartaCodeActionHandler.java
```diff
@@ -178,15 +183,15 @@ public List<CodeAction> codeAction(JakartaJavaCodeActionParams params, IPsiUtils
...
                     }
-//                    if (diagnostic.getCode().getLeft().equals(ManagedBeanConstants.DIAGNOSTIC_CODE_PRODUCES_INJECT)) {
-//                        codeActions.addAll(ConflictProducesInjectQuickFix.getCodeActions(context, diagnostic, monitor));
-//                    }
+                    if (diagnostic.getCode().getLeft().equals(ManagedBeanConstants.DIAGNOSTIC_CODE_PRODUCES_INJECT)) {
+                        codeActions.addAll(ConflictProducesInjectQuickFix.getCodeActions(context, diagnostic));
+                    }
                     if (diagnostic.getCode().getLeft().equals(ManagedBeanConstants.DIAGNOSTIC_CODE_INVALID_INJECT_PARAM)) {
-//                        codeActions.addAll(RemoveInjectAnnotationQuickFix.getCodeActions(context, diagnostic, monitor));
+                        codeActions.addAll(RemoveInjectAnnotationQuickFix.getCodeActions(context, diagnostic));
```
mrglavas, line 190: For each case where more than one quick fix is being used to produce code actions for a specific diagnostic, we're going to need to make sure that each quick fix is working off it's own cloned copy of the context. Maybe not in this PR, but I'd suggest doing a pass over JakartaCodeActionHandler and making an additional copy (or copies) for each diagnostic that has more than one quick fix.

mrglavas (approved):

dshimo (approved):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/JakartaCodeActionHandler.java
turkeylurkey: Good point. The cases up to now have been where one of the pair of quick fixes was making its own copies but I think we should do cloning in this file (even if redundant in some cases) to ensure correctness.

PR #346, opened 2023-04-15
Fix NPE

abhamayee:


src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/operations/codeactions/LSPCodeActionAnnotator.java
```diff
@@ -127,29 +127,31 @@ protected Info doCollectInformationStep(PsiFile file, Editor editor) {
...
-                    params.setTextDocument(new TextDocumentIdentifier(LSPIJUtils.toUri(collectedInfo.file).toString()));
-                    params.setRange(diagnostic.getRange());
-                    CompletableFuture<List<Either<Command, CodeAction>>> codeAction = entry.getKey().getInitializedServer().thenComposeAsync(server -> {
-                        return server.getTextDocumentService().codeAction(params);
-                    });
-                    futures.add(codeAction);
-                    codeAction.thenAcceptAsync(actions -> {
-                        itemInfo.actions = actions;}
-                    );
+        if (collectedInfo != null) {
```
turkeylurkey, line 130: We do not need to call this method if the input is null because `futures` will remain empty. I think this method should be handled the same as `doApplyStep()` at lines 176-179.

turkeylurkey (changes requested):

src/main/java/io/openliberty/tools/intellij/lsp4mp/lsp4ij/operations/codeactions/LSPCodeActionAnnotator.java
abhamayee: Made the suggested changes.

turkeylurkey (approved):

evie-lau (approved):

turkeylurkey:
I merged this pull request because I am not sure if this must be done by a committer.

Thank you for this contribution!

PR #350, opened 2023-04-17
Fix IndexOutOfBoundsException

evie-lau:
mirror redhat-developer/intellij-quarkus#781


dshimo (approved):

evie-lau:
Wondering if the changes to LSPIJUtils.java can be reverted from #320 and #325 because this is now using the correct context.

turkeylurkey:
Actually I think that is a good idea.

PR #351, opened 2023-04-17
Add config and debugger attach tests

mezarin:
This PR:
- Adds tests to process a configuration in Debug mode (toolbar and menu).
- Adds tests that start dev mode using a configuration (toolbar and menu).

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModProjectTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModProjectTestCommon.java
@@ -1,5 +1,15 @@
 package io.openliberty.tools.intellij.it;
 
+/*******************************************************************************
+ * Copyright (c) 2023 IBM Corporation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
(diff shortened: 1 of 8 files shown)


src/test/java/io/openliberty/tools/intellij/it/SingleModProjectTestCommon.java
```diff
@@ -182,277 +207,263 @@ public void testRunTestsActionUsingDropDownMenu() {
...
+                error = null;
+                try {
+                    UIBotTestUtils.importProject(remoteRobot, projectsPath, smMPProjectName);
+                    UIBotTestUtils.openLibertyToolWindow(remoteRobot);
+                    UIBotTestUtils.expandLibertyToolWindowProjectTree(remoteRobot, smMPProjectName);
+                    break;
+                } catch (Exception e) {
+                    // There are instances in which the import actions are successful, but the
+                    // project view never comes up. Instead, the welcome view is shown. If that is the
+                    // case retry.
```
TrevCraw, line 276: do we need to wait before we retry in this scenario?

mezarin: No not really. The lines that could potentially fail have an inherent wait of a number of seconds before they fail. We could add more wait time if we need to, but so far it does not look like we need it.

TrevCraw (approved):

PR #353, opened 2023-04-18
Clean up the creation of code actions

### LINKED ISSUE #310: Quick fixes related to RemoveAnnotationConflictQuickFix

turkeylurkey:
Bring the source compilation unit into the code actions and use to create correct change objects. (Some of these were already done.) Create a sufficient number of copies of the code action context such that each call to getCodeActions() gets a fresh copy of the context. This is necessary to create correct text change proposals.

Fixes #310

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/annotations/PostConstructReturnTypeQuickFix.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/annotations/PostConstructReturnTypeQuickFix.java
@@ -38,7 +38,7 @@ public class PostConstructReturnTypeQuickFix {
         PsiElement node = context.getCoveredNode();
         PsiMethod parentType = getBinding(node);
         String name = "Change return type to void";
-        ChangeCorrectionProposal proposal = new ModifyReturnTypeProposal(name, context.getCompilationUnit(),
+        ChangeCorrectionProposal proposal = new ModifyReturnTypeProposal(name, context.getSource().getCompilationUnit(),
                 context.getASTRoot(), parentType, 0, PsiPrimitiveType.VOID);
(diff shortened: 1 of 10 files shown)


evie-lau (approved):
Looks good to me. I would like to have the massive `if` soup optimized, but that might involve having a map/list of constants (not sure if we currently have one), and can be a later cleanup item.

evie-lau:
Also, now that I see the full code rather than a snippet in Slack, I think `context2` was actually fine 😅 . In case there's ever a future code action that requires a 3rd one, the number naming would be easier to work with.

dshimo (approved):

turkeylurkey:
We can totally clean up that file when we get the time and do it in both the lsp4jakarta and this repository. It grew over time as various people added to it. A good example of the Agile refactoring strategy.

mrglavas:
As future a work item, I honestly think this is worth rewriting to reduce the code duplication. If for instance all of the quick fixes implemented a common interface and the mapping of diagnostics to quick fixes was computed once, most of the logic of codeAction() could be reduced to a table look up (using the diagnostic identifier) and then loop over the associated quick fixes to call getCodeActions() on each.

PR #361, opened 2023-04-20
Run the Jakarta special LS requests on a separate thread.

### LINKED ISSUE #336: [Windows] Slow operation with multiple quick fixes
When a Java file has greater than two diagnostics with quick fixes then the UI becomes slow and choppy. It may hang for several seconds.

To reproduce the issue create a sample project at https://start.microprofile.io/ with MP 3.3, Open Liberty and Java 11. Select all the specifications. Import into IntelliJ and edit ClientController.java in service-a. Add ten new services as shown:
image
Each will be marked with a diagnostic which has the quick fix to initialize it in a constructor. If you hover to show the quick fixes then navigation in the UI becomes choppy and slow. 

We cannot reproduce this on MacOS.

turkeylurkey:
Let me know if adding that method to an Impl class makes sense. 

Fixes #336


evie-lau (approved):
Works much better on Windows now.

And I think it makes sense to move that method to the Impl class. May want to have quarkus also mirror that change.

PR #364, opened 2023-04-21
Config diagnostics

ajm01:
added two tests cases to test diagnostics in server.env and bootstrap.properties

for both testcases:

- insert a valid config name with a known incorrect value into each file
- hover on the flagged incorrect value to provoke a diagnostics popup
- verify the diagnostic string is what it expected.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -708,7 +708,18 @@ public class UIBotTestUtils {
         editor.runJs(String.format(jitterScript, pointX, pointY));
     }
 
-    public static void insertConfigIntoConfigFile(RemoteRobot remoteRobot, String projectName, String fileName, String envCfgNameSnippet, String envCfgNameChooserSnippet, String envCfgValueSnippet) {
+    /**
+     * Inserts a configuration name value pair into a config file via text typing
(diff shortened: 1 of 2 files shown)


src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
```diff
@@ -744,7 +744,7 @@ public static void insertConfigIntoConfigFile(RemoteRobot remoteRobot, String fi
                 "Text " + configNameSnippet + " did not appear in the completion suggestion pop-up window",
```
TrevCraw, line 744: I think `configNameChooserSnippet` would be more valid for lines 743 and 744 instead of `configNameSnippet`. Although it is the combination we're looking for, I think the "chooser" is more of what we are looking for than the original snippet.

src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
```diff
@@ -763,7 +763,7 @@ public static void insertConfigIntoConfigFile(RemoteRobot remoteRobot, String fi
                         "Text " + configNameSnippet + " did not appear in the completion suggestion pop-up window",
```
TrevCraw, line 763: Similar to the other comment, I think `configValueSnippet` should be used on lines 762 and 763.

TrevCraw (approved):

PR #366, opened 2023-04-21
Ensure there is a delay when starting dev mode in debug mode…

### LINKED ISSUE #348: Unable automatically to attach the debugger to JVM
Processing a Liberty configuration in debug mode (Run->debug...) on a slow system results in an error. The error is reported through a message on the IDE and the idea.log.

Associated stack trace:

```
2023-04-15 17:59:02,147 [ 459526]   WARN - #io.openliberty.tools.intellij.util.DebugModeHandler - An error was detected while attaching the debugger to the JVM.
java.lang.Exception: Unable to automatically attach the debugger to JVM on host: localhost and port: 36073. If the debug connection timed out but the server did start successfully, you can still manually create a Remote Java Application debug configuration with the mentioned port and attach to the server.
	at io.openliberty.tools.intellij.util.DebugModeHandler.waitForSocketActivation(DebugModeHandler.java:210)
	at io.openliberty.tools.intellij.util.DebugModeHandler.createAndRunDebugConfiguration(DebugModeHandler.java:133)
	at io.openliberty.tools.intellij.util.DebugModeHandler$1.run(DebugModeHandler.java:118)
	at com.intellij.openapi.progress.impl.CoreProgressManager.startTask(CoreProgressManager.java:425)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.startTask(ProgressManagerImpl.java:114)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcessWithProgressAsynchronously$6(CoreProgressManager.java:476)
	at com.intellij.openapi.progress.impl.ProgressRunner.lambda$submit$3(ProgressRunner.java:252)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$runProcess$2(CoreProgressManager.java:190)
	at com.intellij.openapi.progress.impl.CoreProgressManager.lambda$executeProcessUnderProgress$13(CoreProgressManager.java:591)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:666)
	at com.intellij.openapi.progress.impl.CoreProgressManager.computeUnderProgress(CoreProgressManager.java:622)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:590)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:60)
	at com.intellij.openapi.progress.impl.CoreProgressManager.runProcess(CoreProgressManager.java:177)
	at com.intellij.openapi.progress.impl.ProgressRunner.lambda$submit$4(ProgressRunner.java:252)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1768)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:702)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1$1.run(Executors.java:699)
	at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
	at java.base/java.util.concurrent.Executors$PrivilegedThreadFactory$1.run(Executors.java:699)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

It looks like a wait is missing here:
https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/main/java/io/openliberty/tools/intellij/util/DebugModeHandler.java#L178

turkeylurkey:
… waiting for debugger to connect and reading server.env

Modified the algorithm so it always waits a fixed amount of time waiting for dev mode to start and the Liberty runtime to start. Selected 60 seconds as the timeout. Previously the timeout varied depending on how quickly dev mode started. It varied from a few seconds up to 10 minutes. 

Note that we need to wait for mvn liberty:dev to start, d/l Liberty, unzip it, download and install features and start the Liberty runtime. *Does everyone think this is long enough?*

If the timeout expires a dialog is shown in the IDE.

Fixes #348

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/DebugModeHandler.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/DebugModeHandler.java
@@ -165,12 +165,13 @@ public class DebugModeHandler {
      */
     private String waitForSocketActivation(ProgressIndicator monitor, LibertyModule libertyModule, String host, int debugPort) throws Exception {
         byte[] handshakeString = "JDWP-Handshake".getBytes(StandardCharsets.US_ASCII);
-        int retryLimit = 120; // wait up to 60s for dev mode to start (polling rate 2 per second, 500ms)
-        int envReadMinLimit = 8; // don't check server.env for first 4s
-        int envReadInterval = 4; // recheck server.env every 2s
+        int retryLimit = 360; // wait up to 3 minutes for dev mode to start (polling rate 2 per second, 500ms)
 
         // Retrieve the location of the server.env in the liberty installation at the default location (wpl/usr/servers/<serverName>).
         Path serverEnvPath = getServerEnvPath(libertyModule);
+        // If the server.env has not been created yet then someone did a 'clean' before starting
+        boolean cleanStart = serverEnvPath == null;
+        Path serverEnvBakPath = null;
 
         for (int retryCount = 0; retryCount < retryLimit; retryCount++) {
             // check if cancelled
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/intellij/util/DebugModeHandler.java
```diff
@@ -161,9 +165,9 @@ private void createAndRunDebugConfiguration(ProgressIndicator indicator, Liberty
      */
     private String waitForSocketActivation(ProgressIndicator monitor, LibertyModule libertyModule, String host, int debugPort) throws Exception {
         byte[] handshakeString = "JDWP-Handshake".getBytes(StandardCharsets.US_ASCII);
-        int retryLimit = 300;
-        int envReadMinLimit = 45;
-        int envReadInterval = 5;
+        int retryLimit = 120; // wait up to 60s for dev mode to start (polling rate 2 per second, 500ms)
```
mezarin, line 168: Since attachment seems to be an asynchronous process, we should consider wait times longer than roughly 1 minute. Slow systems along with Gradle built projects with many dependencies may prematurely report that the debugger could not be attached. Increasing the time to wait (i.e. roughly 2/3 minutes) maybe more sensible.

src/main/java/io/openliberty/tools/intellij/util/DebugModeHandler.java
```diff
@@ -177,9 +181,6 @@ private String waitForSocketActivation(ProgressIndicator monitor, LibertyModule
             // The server.env path may not yet exist. If it is null, retry.
             if (serverEnvPath == null) {
                 serverEnvPath = getServerEnvPath(libertyModule);
```
mezarin, line 183: Given that this is an asynchronous process, the removal of the if statement as opposed to just adding a wait to it, now introduces an issue. 
with the changes you made, Let’s say:
- The project has no preexisting server.env
- We detect that there is an available port and decide to use it.
- In the meantime, another process takes the port that we deemed available.
- The LMP/LGP detects the conflict and decides to use a different port. The port is then added to the server.env.
The result is that we will no longer be able to detect the port update, and we will start using the wrong port to connect, which will obviously fail to connect.

The code below it, tries to handle the case where there is a preexisting server.env.

turkeylurkey: When you say "we will no longer be able to detect the port update" do you mean "we are not able to detect the port update and never were?" The old code looks like it loops until server.env is created and then expects server.env.bak to be there.

mezarin: By this: "we will no longer be able to detect the port update" i mean that the wait method "quickly" finds a random port and exits to then try to connect the debugger to that port. After the method exits, we no longer check for a port update.

The old code + a wait in the if statement would have waited for the LGP/LMP to create the server.env. At that point,  we know that it is ok to check the serve.env for a debug port entry. 

Now, If there were a preexisting server.env, the code would wait for serer.env.bak to be present to check for the port in server.env because LGP/LMP will only create the .bak file when it is updated.
Having said that, you can also argue that the old code does not completely close all issues, but the intent i believe is to close the timing windows as much as possible.

The alternative would be to avoid checking for server.env updates altogether with the assumption that picking a random port is good enough as it will probably work 99% of the time. The old code tries to solve 1/2 the 1% involving server.env because the current assumption is that server.env is located at the default location. The other 1/2 of the solution involves detecting a custom server.env location and also fixing the issue i described above.

turkeylurkey: Thank you for explaining why we check for the .bak file but I'm not sure it helps. The way it works now is that it just tries to connect to the random port a few times and if that does not work it checks the server.env for an update. (I guess in reality dev mode will always create the server.env before it starts the Liberty process.) So this assumes that if the port is already in use the attempt to connect will fail (not always true). However I'm not sure we can guarantee that this code will run before dev mode starts (but probably usually does). 
I'm not sure the .bak file helps because we could be in a fresh directory (after a mvn clean) so there is a server.env and no .bak so we just use the server.env. Or we could be in a restart situation in which case dev mode makes a server.env and the .bak file. If we do not know how long dev mode takes to start how do we know how long to wait to see if a .bak file will appear?
The old code tries multiple times to get the name of the server.env, is this necessary?

turkeylurkey: I guess we can almost guarantee that this code runs before dev mode starts. I think you are saying always to check server.env before trying to attach. I agree, this is probably safest. Please see the latest rewrite.

turkeylurkey: @TrevCraw @mezarin I tried two loops: one polling slowly while dev mode starts and then one polling quicker while liberty starts but this failed in the case where dev mode changes the debug port due to a collision. 

We can stick with one loop that only polls once every 5s because this whole process takes a while and faster polling does not help much. This way the file I/O is reduced and the user only needs to wait an extra 2s on average.

I suspect this fix is needed for testing so I think we should merge it and open another issue if required to address the issue where the server.env is in a custom location.

TrevCraw (approved):

Issue #328, opened 2023-04-04
[LSP4MP] Diagnostics regarding properties file do not update instantly


evie-lau:
As we share a codebase with [intellij-quarkus](https://github.com/redhat-developer/intellij-quarkus), this issue is present there too. The VSCode extension `Tools for MicroProfile` does not have this same behavior.

When updating a MP config properties file, it can take up to a minute for the changes to be reflected in memory. During this time, any diagnostics refresh on the java file referencing the properties file will be using an outdated version of the properties file.

Even when the memory copy of the properties file gets updated, `publishDiagnostics` isn't triggered automatically, and something else has to trigger it to update the diagnostics and refresh the problems.

This is especially noticeable when using a quickfix to add a property into the properties file, because the diagnostic will stick around until after the system file gets re-read, and the diagnostics are triggered again.

In summary: to have the MP diagnostics accurately reflect a changed properties file, you must...

1. wait up to 60 seconds for the in-memory copy to read from the file in system
2. manually trigger a `publishDiagnostic` on the java file you are viewing (changing or closing/reopening the file)

mrglavas:
@evie-lau Do you happen to have a test case I could use to reproduce this?

evie-lau:
After some more debugging with @mrglavas, here are some things we observed.
Update: it is not a timing issue... it requires switching windows on the OS.

When applying a quickfix or manual change to modify microprofile-config.properties:
- nothing will ever evict MP config cache until switching windows
- nothing will ever loadConfigSource until making changes to the java file, AFTER the config cache has been evicted

loadConfigSource is how we pull in the contents of the config file, but it only runs when the cache has been evicted (is null). The cache does not get nulled until switching out of the IDE window, so diagnostics will never have an updated copy of the config file to work with until this happens.

Evict call comes from MicroProfileProjectListener (implements ModuleListener and BulkFileListener)
loadConfigSources comes from PsiMicroProfileProject

mrglavas:
@evie-lau When we were debugging yesterday were you saving the files? The response to this post on JetBrains' forums: https://intellij-support.jetbrains.com/hc/en-us/community/posts/360006715420-Virtual-File-changes-only-occur-when-clicking-off-Intellij suggests that events are only fired when changes are saved on disk. Merely typing in the editor isn't enough. Perhaps the quickfix for `microprofile-config.properties` is only queuing up a file write and switching windows forces the in memory document to be written to the disk.

[...]

PR #373, opened 2023-04-27
Listen to in-memory document updates to make the cache more responsive to config changes.

mrglavas:
Required to address #328.

This PR also fixes several NPEs that could occur if cache eviction overlaps with other operations on the PsiMicroProfileProject. The way PsiMicroProfileProject was written probably still isn't thread-safe, but this addresses the more obvious race conditions.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/project/AbstractConfigSource.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/project/AbstractConfigSource.java
@@ -10,6 +10,8 @@
 package io.openliberty.tools.intellij.lsp4mp4ij.psi.core.project;
 
 import com.intellij.openapi.compiler.CompilerPaths;
+import com.intellij.openapi.editor.Document;
+import com.intellij.openapi.fileEditor.FileDocumentManager;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.vfs.VfsUtil;
@@ -19,6 +21,8 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.io.Reader;
+import java.io.StringReader;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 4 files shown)


evie-lau:
Tested on commit [c9f360a](https://github.com/OpenLiberty/liberty-tools-intellij/pull/373/commits/c9f360ab10e2a4e34b9559bf25454265e6da158f) just now, but not seeing any change in behavior yet - still not loading the new mpconfig file until after switching windows.

mrglavas:
@evie-lau Some other component has to call PsiMicroProfileProject.getConfigSources() to cause the config file to be reloaded. Try adding a print statement around evictConfigSourcesCache(). I see the method being called when the quick fix is applied or if I manually type characters into the editor for the config file.

mrglavas:
@evie-lau It might be the case that the properties are being loaded from the file on disk rather than the live document in the editor. If that is the case then that is yet another problem.

evie-lau:
Hmm, okay, when using quickfix/manual edits, I do see the cache is evicted and it is loading the config file again (although after accepting diagnostics). 
But for some reason, subsequent diagnostics updates still aren't reflecting the changed config file until switching windows and back.

mrglavas:
I've updated the PR to load the properties file from the Document. In my testing, now when I apply the quick fix, I can get the diagnostic to be refreshed by merely touching the Java source file in the editor. Also observing this working with manual changes to the properties file. No explicit file saves required.

evie-lau:
@angelozerr curious what you think of these changes, and whether these should be made to the base LSP4IJ code. Also not 100% sure if these show up in quarkus-intellij.

To summarize, we noticed that when modifying the `microprofile-config.properties` (either quickfix or manual change) to fix a diagnostic (ie `@ConfigProperty`), to make the diagnostics update correctly we need to do the following steps:
1) switch windows outside the IDE (this writes the `microprofile-config.properties` to disk) OR manually save the file (not made obvious with IntelliJ UI)
2) make a change to the Java file with the `@ConfigProperty` annotation to trigger the LSP diagnostics to read the new file from disk

This PR fixes problem 1, but not 2. Instead of reading from the hard disk file which requires manual actions, we read from the in-memory document. 
IntelliJ seems to tout "autosaving" as a feature, so we can't expect users to manually save (in step 1), especially when its UI does not have an indicator for unsaved files. So this change is a step towards the MP diagnostics updating correctly reflecting config file changes.

angelozerr:
@evie-lau to be honnest with you, I'm newbie with IJ and for the moment my main goal is to provide a robust LSP client for IJ (my skills are more about LSP since I have created and worked on several language server like LemMinx, LSP4MP and I contributed to  LSP4E, LSP client for Eclipse IDE).

Our second goal is to align last feature that we implemented in LSP4MP with @datho7561 in IJ. We will try to investigate about your issue and we will able to give you (I hope) relevant feedback.

Thanks for asking me and I'm sorry I can't help you now.

evie-lau (approved):
Looks good, seeing the reloading of the live document when triggering diagnostics without having to manually save the file or switch windows at the OS level.

angelozerr:
I will have intention to fix this issue in IJ Quarkus
https://github.com/redhat-developer/intellij-quarkus/issues/860

If I remember in vscode we need to save the properties file to refresh Java diagnostics.

If I understand your fix you refresh java diagnostics when you are updating properties file? If it that it is not good for performance. It is like Java in JDT, it refresh diagnostics for other Java file when you save the file.

PR #383, opened 2023-05-02
Add handling for variable substitution in the form ${ABC}

### LINKED ISSUE #372: LSP4MP Completion: substitute correct class name in template
When you use the completion item "mpliveness" a snippet of code is inserted but the name of the new class is not replaced properly in the template.
image
In this image we see "${TM_FILENAME_BASE}" where we would expect the IDE to insert the correct class name.

turkeylurkey:
Added the code in the style currently in use but it could stand a rewrite.

Fixes #372


evie-lau (approved):
Agreed with potential rewrite in the future. Very hard to read a method ~150 lines long.

PR #411, opened 2023-05-17
Ensure getCodeActions() does not return null and returns a list instead.

turkeylurkey:
This mirrors a fix in lsp4jakarta. In very short methods I changed the return to Collections.emptylist() but in the slightly longer methods I moved the definition of the return value out of the `if`. This is more like what we do in lsp4jakarta JDT component.


mrglavas (approved):
Looks good to me.

evie-lau (approved):

PR #421, opened 2023-05-23
return to getLineStartOffset. catch OoB

dshimo:


turkeylurkey:
I've been looking at this today and there is a problem where we use the same document for source and target. So it is using the target document to calculate the range. I'm just testing a fix.

dshimo:
Closing in favor of #422

PR #438, opened 2023-06-06
Close the error dialog if the debugger failed to attach

mezarin:
This PR will prevent a cascading effect of failures if a test exercising the debugger attachment fails with an error dialog that takes focus away from the main frame.


ajm01 (approved):
LGTM

PR #446, opened 2023-06-13
fix intermittent file open issue in mpls tests

ajm01:


mezarin (approved):
Based on the decision made to disable all LS tests on Windows, the updates look good.

Issue #445, opened 2023-06-12
Liberty Dashboard still detects Gradle project when plugin is commented


evie-lau:
When the plugin definition for Liberty Gradle Tools `io.openliberty.tools.gradle.Liberty` is commented, the Liberty dashboard still detects the project. 
<img width="1138" alt="image" src="https://github.com/OpenLiberty/liberty-tools-intellij/assets/689163/334eaeb6-f0fe-4403-8edc-f32d572395fd">

May be due to the regex used to detect the plugin definition.

Could rework the regex, or preferably programmatically read the plugin IDs to check for it instead, if possible.

dshimo:
Seems like the [negative lookbehind](https://github.com/OpenLiberty/liberty-tools-intellij/blob/63dc3c6d408810a15e2b31051658896b38d9c342/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java#L100) for comments simply didn't work.

turkeylurkey:
If you want to stick with negative lookbehind then "(?<!//\s*)" should do it.

PR #448, opened 2023-06-13
tiny fix for comment regex

dshimo:
Fix for #445


turkeylurkey (approved):
Just need to verify when we can merge.

PR #450, opened 2023-06-14
Update readme with JAVA_HOME for choosing JDK to run Liberty

evie-lau:


turkeylurkey (approved):

docs/user-guide.md
```diff
@@ -26,7 +26,7 @@ This guide provides detailed instructions on how to import and configure your Li
 ## Before you begin
 ### Software requirements
 - **IntelliJ IDEA**: Starting with the Liberty Tools for IntelliJ IDEA 0.0.8 release, IntelliJ IDEA version 2022.2 is required. Liberty Tools for IntelliJ IDEA is compatible with the Community Edition of IntelliJ IDEA.
-- Java 17 is required by Liberty Tools for IntelliJ IDEA and bundled with IntelliJ IDEA version 2022.2+. If you have changed the IntelliJ IDEA Boot Java Runtime through the "Change Boot Java Runtime for the IDE" preference, ensure that you are using Java 17.
+- Java 17 is required by Liberty Tools for IntelliJ IDEA and bundled with IntelliJ IDEA version 2022.2+. If you have changed the IntelliJ IDEA Boot Java Runtime through the "Change Boot Java Runtime for the IDE" preference, ensure that you are using Java 17 or later.
```
TrevCraw, line 29: wondering if we should say "Java 17 or later" at the start of this sentence as well

turkeylurkey: I think once is enough.

evie-lau:
Realized this changed text is duplicated in both the main README and the user-guide.md. Will be mirroring the change to the other file that I missed in the first commit.

docs/user-guide.md
dmuelle, line 29:
```suggestion
- Liberty Tools for IntelliJ IDEA requires a minimum of Java 17, which is bundled with IntelliJ IDEA version 2022.2+. If you change the IntelliJ IDEA Boot Java Runtime through the **Change Boot Java Runtime for the IDE** preference, ensure that you are using Java 17 or later.
```

docs/user-guide.md
```diff
@@ -41,9 +41,14 @@ This guide provides detailed instructions on how to import and configure your Li
 
 ### Settings
 
-Liberty Tools for IntelliJ IDEA honours  the following external settings:
-- Maven home path set in the **Preferences > Build, Execution, Deployment > Build Tools > Maven** window when running Liberty dev mode on Maven projects.
-- Gradle JVM set in the **Preferences > Build, Execution, Deployment > Build Tools > Gradle** window when running Liberty dev mode on Gradle projects.
+Liberty Tools for IntelliJ IDEA honors the following settings:
+
+#### IntelliJ settings:
+- Maven home path set in the **Settings > Build, Execution, Deployment > Build Tools > Maven** window when running Liberty dev mode on Maven projects.
```
dmuelle, line 47:
```suggestion
- The Maven home path that is set in the **Settings > Build, Execution, Deployment > Build Tools > Maven** window when you run Liberty dev mode on Maven projects.
```

docs/user-guide.md
```diff
@@ -41,9 +41,14 @@ This guide provides detailed instructions on how to import and configure your Li
...
+- Gradle JVM set in the **Settings > Build, Execution, Deployment > Build Tools > Gradle** window when running Liberty dev mode on Gradle projects.
```
dmuelle, line 48:
```suggestion
- The Gradle JVM that is set in the **Settings > Build, Execution, Deployment > Build Tools > Gradle** window when you run Liberty dev mode on Gradle projects.
```

docs/user-guide.md
```diff
@@ -41,9 +41,14 @@ This guide provides detailed instructions on how to import and configure your Li
...
+
+#### Environment Variables:
+- **JAVA_HOME** in the current terminal for running Liberty.
```
dmuelle, line 51: what does "running liberty" mean in this context? Does it mean the **JAVA_HOME** value that is set in a `server.env` file for a given instance of Liberty? The answer might be obvious for the user but this seems a bit vague

dmuelle (changes requested):
ID review

docs/user-guide.md
evie-lau: Ah, to be more clear, the **JAVA_HOME** system environment variable (not server.env) is used to specify which Java to use when running Liberty dev mode.

docs/user-guide.md
TrevCraw: @evie-lau I think we should mention the Gradle "Distribution" setting instead, which I believe is what LT hounours... but we should double check

evie-lau: It does indeed honor the Distribution setting. 
But for the Gradle JVM setting just below it, I don't notice any difference when changing that.

docs/user-guide.md
```diff
@@ -37,13 +37,18 @@ This guide provides detailed instructions on how to import and configure your Li
...
 ### Settings
 
-Liberty Tools for IntelliJ IDEA honours  the following external settings:
-- Maven home path set in the **Preferences > Build, Execution, Deployment > Build Tools > Maven** window when running Liberty dev mode on Maven projects.
-- Gradle JVM set in the **Preferences > Build, Execution, Deployment > Build Tools > Gradle** window when running Liberty dev mode on Gradle projects.
+Liberty Tools for IntelliJ IDEA honors the following settings:
+
+#### IntelliJ settings:
+- The Maven home path that is set in the **Settings > Build, Execution, Deployment > Build Tools > Maven** window when you run Liberty dev mode on Maven projects.
+- The Gradle Distribution that is set in the **Settings > Build, Execution, Deployment > Build Tools > Gradle** window when you run Liberty dev mode on Gradle projects.
```
evie-lau, line 48: I noticed that depending on the IntelliJ UI (Classic vs New), it's called either Preferences or Settings. Should we make a note of that?
Not sure if just calling it **Settings/Preferences** is ok, or maybe stick with **Preferences** and add a note saying it may be called **Settings** if using the new UI.

docs/user-guide.md
```diff
@@ -37,13 +37,18 @@ This guide provides detailed instructions on how to import and configure your Li
 
     - Liberty Maven Plugin -> 3.7.1
 
-    - Liberty Gradle Plugin -> 3.5.1
```
evie-lau, line 40: Noting that this change to recommended minimum is to support using Gradle 8.x, which did not work with LGP 3.5.x

docs/user-guide.md
```diff
@@ -37,13 +37,18 @@ This guide provides detailed instructions on how to import and configure your Li
...
+    - Liberty Gradle Plugin -> 3.6.1
```
evie-lau, line 40: The Gradle JVM that is set in the **Settings > Build, Execution, Deployment > Build Tools > Gradle** window when you run Liberty dev mode on Gradle projects.

evie-lau, line 40: Noting that this change for the recommended minimum is to support Gradle 8.x, which LGP 3.5.x did not support.

docs/user-guide.md
dmuelle: @evie-lau what are the two views called w/in Intellij- Classic and New? Is that version-dependent or an option where the user chooses which UI to use, regardless of their version?

docs/user-guide.md
dshimo: Nice catch. I'll also upgrade LMP to 3.8.1

docs/user-guide.md
```diff
@@ -37,13 +37,18 @@ This guide provides detailed instructions on how to import and configure your Li
...
-- Maven home path set in the **Preferences > Build, Execution, Deployment > Build Tools > Maven** window when running Liberty dev mode on Maven projects.
-- Gradle JVM set in the **Preferences > Build, Execution, Deployment > Build Tools > Gradle** window when running Liberty dev mode on Gradle projects.
+Liberty Tools for IntelliJ IDEA honors the following settings:
+
+#### IntelliJ settings:
+- The Maven home path that is set in the **Settings > Build, Execution, Deployment > Build Tools > Maven** window when you run Liberty dev mode on Maven projects.
+- The Gradle Distribution that is set in the **Settings > Build, Execution, Deployment > Build Tools > Gradle** window when you run Liberty dev mode on Gradle projects.
+
+#### Environment Variables:
+- The **JAVA_HOME** system environment variable in the current terminal is used when you run Liberty dev mode in any project.
```
evie-lau, line 51: @dmuelle Updated to clarify this. Let me know if there should be any wording changes.

PR #486, opened 2023-09-05
Fixes #173: Updated the regex to strip the Libery Config LS jars version

### LINKED ISSUE #173: Strip Liberty Config LS jar version when copying over into build/server directory
In the build.gradle, we strip the versions of the LS jars so that we don't have to update the launcher code for new versions. 
https://github.com/OpenLiberty/liberty-tools-intellij/blob/345cb9c2f3168a23e685f823351e17a4894f5822/build.gradle#L72

The version is still present for the Liberty Config LS jars:
image
https://github.com/OpenLiberty/liberty-tools-intellij/blob/9dd865598b22e75111ae9fe206c910a8292440bb/src/main/java/io/openliberty/tools/intellij/liberty/lsp/LibertyConfigLanguageServer.java#L35

dessina-devasia:
Fixes #173 
Updated the regex to strip the Libery Config LS jars version

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -121,7 +121,7 @@ dependencies {
 task copyDeps(type: Copy) {
     from configurations.lsp
     into new File(buildDir, 'server/server')
-    rename '^(.*)(-[0-9]+.[0-9]+.[0-9]+(-SNAPSHOT)?)(.*)$', '$1$4'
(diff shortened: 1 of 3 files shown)


TrevCraw (approved):

TrevCraw:
It does not look like the test failures for the Windows build are related to the changes in this PR. This PR should be safe to merge. Opened a new issue to investigate the test failures further: https://github.com/OpenLiberty/liberty-tools-intellij/issues/494

PR #502, opened 2023-09-21
Transform the Jakarta EE diagnostic collectors on to a common extension point shared with MicroProfile.

### LINKED ISSUE #501: Transform the Jakarta EE diagnostic collectors on to a common extension point shared with MicroProfile.
- Update the diagnostic participant extension point so that it can be used by any Java language server (MicroProfile, Jakarta EE, etc...).
- Refactor the code for loading and invoking diagnostic participants so that it can be shared with MicroProfile and Jakarta EE.
- Replace the diagnostics code in the Jakarta EE properties manager with calls to the shared/refactored code.
- Refactor all of the Jakarta EE diagnostic collectors to implement the common extension point.

Tests for Jakarta EE diagnostics were delivered with the 23.0.6 release and can be used for regression testing.

mrglavas:
Resolves #501.

1. Transformed `AbstractDiagnosticsCollector` so that it implements `IJavaDiagnosticsParticipant`. This is the base class for all of the Jakarta EE diagnostic collectors.
2. Created `JavaDiagnosticsDefinition`, a bean wrapper class around `IJavaDiagnosticsParticipant` which includes a group field to distinguish between diagnostics for different language servers (e.g. MP vs. Jakarta EE).
3. Refactored diagnostics related code from the MP `PropertiesManagerForJava` into its own class: `DiagnosticsHandler`, enhancing this code to use the group field when filtering the list of diagnostics collectors.
4. Plugged the `DiagnosticsHandler` into the `PropertiesManagerForJakarta`, replacing the diagnostics related code from this class.
5. Changed the definition of the `javaDiagnosticsParticipant` in the plugin.xml to be the `JavaDiagnosticsDefinition` bean class, allowing a group attribute to be specified for each of the diagnostics collectors.
6. Specified a `javaDiagnosticsParticipant` in the plugin.xml for each of the Jakarta EE diagnostics and updated all of the existing `javaDiagnosticsParticipant`s to specify that those diagnostics apply to MicroProfile by adding the group field.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AbstractDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AbstractDiagnosticsCollector.java
@@ -13,12 +13,16 @@
 
 package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import com.intellij.psi.*;
 import com.intellij.psi.search.GlobalSearchScope;
+import io.openliberty.tools.intellij.lsp4mp4ij.psi.core.java.diagnostics.IJavaDiagnosticsParticipant;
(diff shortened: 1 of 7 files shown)


turkeylurkey (approved):
Looks good to me.

TrevCraw (approved):

PR #503, opened 2023-09-22
Part of #198 Added code to shift focus to the terminal

### LINKED ISSUE #198: Terminal focus doesn't shift and terminal renaming doesn't work
When we run on a Liberty action on a project, the focus is not automatically pulled to the corresponding terminal window for the project (if it was not already there) even though we set visible and request focus:
https://github.com/OpenLiberty/liberty-tools-intellij/blob/2908d53a16704abbb812a4afb6095e9e3907d58e/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java#L182-L183

Also noting that terminal renaming doesn't work, issue open against JetBrains:  https://youtrack.jetbrains.com/issue/IDEA-297207

See https://github.com/OpenLiberty/liberty-tools-intellij/issues/143#issuecomment-1319348592

Blocked on fixing this until we have the API fixes from Jetbrains

vaisakhkannan:
Part of #198
Added code to shift focus to the terminal

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches: '**'
   pull_request:
-    branches: [ main ]
+    branches: [ terminal-focus-fix-branch ]
 
 jobs:
   build:
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2022 IBM Corporation.
+ * Copyright (c) 2020, 2023 IBM Corporation.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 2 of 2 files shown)


vaisakhkannan:
I added some code in _LibertyProjectUtil.java_. It's actually working, but there are two exceptional cases:

1. If two projects have the same name, let's say "**_example_**" and I perform a Liberty action, it will open a terminal window named "_**example**_". If I then perform a Liberty action in another project with the same name, "**_example,_**" it will open a terminal session named "_**example (1)**_". However, when I attempt to perform a Liberty action on the "**_example (1)_**" project, the focus always reverts to the "**_example_**" project.

 (Note: Both project names in the Liberty dashboard remain as "**_example_**")
 
2.  When I open the project “**_guide-getting-started_**", the terminal tab name gets stripped, and the terminal window name appears as '**_guide-get…ng-started_**' .When I check the condition "_**widgetName.equals(displayName)**_", it fails. The reason behind this failure is that the **widgetName** is '**_guide-getting-started_**' and the displayName is '**_guide-get…ng-started_**' , so they don't match. This mismatch causes the focus to shift to the **_Local_** tab named terminal session.

TrevCraw (approved):

vaisakhkannan:
Closing this pull request since the original pull request at https://github.com/OpenLiberty/liberty-tools-intellij/pull/637 has been merged.

PR #524, opened 2023-10-09
Fixes #167 Custom Start... added to select Run configurations

### LINKED ISSUE #167: Prompt selection when there are multiple Liberty run configs and Start... is selected
Currently when Start... is selected, if >1 Liberty run config exists, the currently selected run config in the run config view is chosen. If the currently selected run config does not map to the same Liberty module, the first run config in the list will be chosen. 
Enhancement will be to allow the user to choose between the valid run configs.
image

Rahul-Biju:
Fixes #167 - Custom "Start..." added to select Run configurations.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyDevCustomStartAction.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyDevCustomStartAction.java
@@ -18,9 +18,11 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VfsUtil;
 import com.intellij.openapi.vfs.VirtualFile;
 import io.openliberty.tools.intellij.LibertyModule;
+import io.openliberty.tools.intellij.LibertyPluginIcons;
 import io.openliberty.tools.intellij.runConfiguration.LibertyRunConfiguration;
 import io.openliberty.tools.intellij.runConfiguration.LibertyRunConfigurationType;
 import io.openliberty.tools.intellij.util.LocalizedResourceUtil;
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/intellij/actions/LibertyDevCustomStartAction.java
```diff
@@ -63,23 +65,29 @@ protected void executeLibertyAction(LibertyModule libertyModule) {
...
             selectedLibertyConfig = createNewLibertyRunConfig(runManager, libertyModule);
         } else {
-            // TODO if 1+ run configs, prompt user to select the one they want see https://github.com/OpenLiberty/liberty-tools-intellij/issues/167
             // 1+ run configs found for the given project
-            RunnerAndConfigurationSettings selectedConfig = runManager.getSelectedConfiguration();
-            if (libertyModuleSettings.contains(selectedConfig)) {
-                // if the selected config is for the Liberty module, use that run config
-                selectedLibertyConfig = selectedConfig;
+            final String[] runConfigNames = toRunConfigNames(libertyModuleSettings);
+            final String[] runConfigNamesTooltips = toRunConfigNamesTooltips(libertyModuleSettings);
```
aparnamichael, line 70:
```suggestion
            final String[] runConfigNamesTooltips = new String[libertyModuleSettingssize())];
```
Can we pass an empty array in LibertyProjectChooserDialog constructor? I think if we pass like that there won't be any tooltips. As per current code we are showing not showing any additional code in tooltip.

Rahul-Biju:
incorrectly opened - not created from my fork.

PR #525, opened 2023-10-11
Resolve functional issues with ExtendClassProposal.

### LINKED ISSUE #520: Resolve functional issues with LSP4Jakarta's ExtendClassProposal.
There are a couple of problems with the current implementation of `io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction.proposal.ExtendClassProposal`.

1. The target `PsiFile` is not an argument in the constructor. A value of null is passed to the super constructor in its place. This will always cause a `NullPointerException` to be thrown when `performUpdate()` is called.
2. `ExtendClassProposal` extends `ImplementInterfaceProposal` which is hard-wired for adding interfaces to a class. Applying this for class inheritance is wrong (implements vs. extends) and will generate source that will not compile.

Resolving both of these issues is required in order to complete the work on #517.

mrglavas:
Resolves #520. See the detailed description of the problem on this issue.


turkeylurkey (approved):
👍

aparnamichael (approved):
Code changes looks fine for me.

Rahul-Biju:
Code fine for me. No issues.

PR #527, opened 2023-10-13
Fixes #222 Added hover info to entries in the Liberty Tool window

### LINKED ISSUE #222: Add hover info to entries in the Liberty Tool window

Rahul-Biju:
Fixes #222 Added hover info to entries in the Liberty Tool window


Rahul-Biju:
incorrectly opened - not created from my fork.

PR #528, opened 2023-10-13
Fixes #222 add hover info to entries in the Liberty Tool window

### LINKED ISSUE #222: Add hover info to entries in the Liberty Tool window

Rahul-Biju:
Fixes #222 Added hover info to entries in the Liberty Tool window

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
+++ b/src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
@@ -305,8 +305,11 @@ public class LibertyExplorer extends SimpleToolWindowPanel {
                 boolean leaf,
                 int row,
                 boolean hasFocus) {
+            /**
+             * modification to get tooltiptext for liberty module tree
+             */
+            tree.setToolTipText(getBuildPath(value));
             super.getTreeCellRendererComponent(tree, value, sel, expanded, leaf, row, hasFocus);
-
             // assign gear icon to action nodes
             if (leaf) {
                 setIcon(LibertyPluginIcons.IntelliJGear);
@@ -325,6 +328,11 @@ public class LibertyExplorer extends SimpleToolWindowPanel {
 
             return this;
         }
+    }
+    
+    private static String getBuildPath(Object value) {
+        LibertyModuleNode node = (LibertyModuleNode) value;
+        return node.getFilePath().toString().replace("file://", "");
     }
 
     private static void executeAction(Tree tree) {


aparnamichael (approved):
Looks good to me.

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -220,6 +222,33 @@ public static Tree buildTree(Project project, Color backgroundColor) {
...
+                TreePath path = tree.getPathForLocation(e.getX(), e.getY());
+                if (path == null ) {
+                    if (currentTooltipText != null) {
+                        tree.setToolTipText(null);
+                        currentTooltipText = null;
+                    }
+                } else {
+                    Object node = path.getLastPathComponent();
+                    if (node instanceof LibertyModuleNode) {
+                        LibertyModuleNode treeNode = (LibertyModuleNode) node;
```
TrevCraw, line 241: recommend renaming "treeNode" to "libertyNode" for clarity

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -220,6 +222,33 @@ public static Tree buildTree(Project project, Color backgroundColor) {
...
+                        String tooltipText = getBuildPath(treeNode);
```
TrevCraw, line 242: Instead of using a new method here, it is best to use available ones. It looks like your new `getBuildPath()` method was written to get the build file, convert it to a String and then replace part of the String to get the correct file path. 

It would be simpler to call the LibertyModuleNode method `getFilePath()` from here, which retrieves a VirtualFile object (this method is confusingly named since it does NOT return a file path, but a file object), and then with the VirtualFile object, you can call `getPath()` to retrieve the file path in a String format to then display to the user.

For example:

```
VirtualFile buildFile = libertyNode.getFilePath();
String filePath = buildFile.getPath();
String tooltipText = filePath;
```
or more condensed:
```
String tooltipText = libertyNode.getFilePath().getPath();
```

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -326,6 +355,14 @@ public Component getTreeCellRendererComponent(
             return this;
         }
     }
+    
+    private static String getBuildPath(Object value) {
```
TrevCraw, line 359: This method will not be needed after making the suggested changes above.

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -32,12 +32,14 @@
...
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.tree.DefaultTreeCellRenderer;
+import javax.swing.tree.TreeCellRenderer;
 import javax.swing.tree.TreePath;
 import javax.xml.parsers.ParserConfigurationException;
 import java.awt.*;
 import java.awt.event.KeyAdapter;
 import java.awt.event.KeyEvent;
 import java.awt.event.MouseEvent;
+import java.awt.event.*;
```
TrevCraw, line 42: Instead of importing all `java.awt.event` packages, it would be better to just import the one that is needed here, which is `java.awt.event.MouseMotionAdapter`

TrevCraw (changes requested):

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -32,12 +32,14 @@
 import javax.swing.*;
 import javax.swing.tree.DefaultMutableTreeNode;
 import javax.swing.tree.DefaultTreeCellRenderer;
+import javax.swing.tree.TreeCellRenderer;
```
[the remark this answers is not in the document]
Rahul-Biju: done.

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
Rahul-Biju: done.

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
Rahul-Biju: done.

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
Rahul-Biju: done.

TrevCraw (changes requested):
The tooltip text covers up the Liberty actions and other project nodes when moving the mouse off of the project node. This should be addressed before merging these changes.

https://github.com/OpenLiberty/liberty-tools-intellij/assets/5885743/2adcaa55-0476-4f19-8317-62efd4a21b84

PR #573, opened 2023-11-09
Use the Gradle config in a subdirectory or default to that in the Project directory.

### LINKED ISSUE #568: FVT: Opening a parent folder with two gradle projects not trigger the "Start action."
In Linux, when you try to open a parent folder containing two gradle projects, the "Start action" is not triggered.

I tried Using the below Configurations
OS : RedHat Linux
Version : 9.2
Intellij Version: 2023.2.4

Screenshot 2023-11-07 at 10 13 10 PM

Screenshot 2023-11-07 at 2 59 23 PM

This issue is related to #563 , which specifies Maven project. However, in this case, it's  Gradle project.

 I've tried it using both Wrapper and Local Installation in the Gradle Distribution of Build Tool. I am encountering the same popup message in both cases when trying to start the server.

turkeylurkey:
Fixes #568

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
@@ -183,10 +183,6 @@ public class LibertyGradleUtil {
     public static String getGradleSettingsCmd(Project project, VirtualFile buildFile) throws LibertyException {
         GradleProjectSettings gradleProjectSettings = GradleSettings.getInstance(project).getLinkedProjectSettings(buildFile.getParent().getPath());
         if (gradleProjectSettings == null) {
-            // Try the project dir after trying the directory containing the build file assuming it was a subdirectory in the project.
-            gradleProjectSettings = GradleSettings.getInstance(project).getLinkedProjectSettings(project.getBasePath());
-        }
-        if (gradleProjectSettings == null) {
             String translatedMessage = LocalizedResourceUtil.getMessage("gradle.settings.is.null");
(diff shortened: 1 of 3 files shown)


dessina-devasia (approved):

src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
```diff
@@ -206,13 +206,13 @@ else if (DistributionType.LOCAL.equals(gradleProjectSettings.getDistributionType
...
-     * @param project liberty project
+     * @param wrapperDir location of gradle wrapper
      * @return the Graddle wrapper path to be executed or an exception to display
      * @throws LibertyException
      */
-    private static String getLocalGradleWrapperPath(Project project) throws LibertyException {
+    private static String getLocalGradleWrapperPath(String wrapperDir) throws LibertyException {
         String gradlew = SystemInfo.isWindows ? ".\\gradlew.bat" : "./gradlew";
-        File file = new File(project.getBasePath(), gradlew);
+        File file = new File(wrapperDir, gradlew);
```
TrevCraw, line 215: do you want to make the same change here as on the Maven side and change the name `file` to `wrapper`?

TrevCraw: actually, let's leave this for now. We can update the naming to be consistent with Maven another time.

turkeylurkey: Good catch!

TrevCraw (approved):

Issue #506, opened 2023-09-26
Improper Terminal focus configuration for a project with the long project name


vaisakhkannan:
When dealing with a lengthy project name, such as "**guide-getting-started**", I observed a change or alteration in the terminal tab's name to '**guide-get…ng-started**'. While addressing issue #198, I made the following observation, when attempting to select a terminal session from the terminal tool window, we are presented with the following properties:

- Content name: guide-get…ng-started
- Tab name: 'guide-get…ng-started'
- Tool window: 'guide-get…ng-started'

It is worth noting that all of these properties display a modified name. Consequently, when attempting to set the focus by comparing the widget name with these altered names, we encounter difficulties in identifying the correct terminal window.

vaisakhkannan:
With the update in the Maven-test-framework and IntelliJ version, I observed that the above issue has been resolved.
Here, we can see there is no alteration of name for lengthy project name.

<img width="1155" alt="Screenshot 2024-01-30 at 11 26 45 AM" src="https://github.com/OpenLiberty/liberty-tools-intellij/assets/133869023/1d6cd0a1-ad2a-48db-8c28-5150754c8255">

vaisakhkannan:
There is a PR (https://github.com/OpenLiberty/liberty-tools-intellij/pull/637) open for the common issue (https://github.com/OpenLiberty/liberty-tools-intellij/issues/198). Once the above PR is merged, we can close this issue.

vaisakhkannan:
The PR https://github.com/OpenLiberty/liberty-tools-intellij/pull/637 has been merged, and the problem described in this issue has also been addressed with the same PR. I am now closing this issue.

Issue #505, opened 2023-09-26
Improper Terminal focus configuration for a project with the same name


vaisakhkannan:
When multiple projects share the same name, such as "**_sample_**", new terminal sessions are generated with names like "sample," "sample (1)," "sample (2)," and so on. All of these terminal sessions retain the widget name "**_sample_**".

During investigation of issue #198, I discovered that, when attempting to perform a Liberty action on the "**_sample (1)_**" project, the focus consistently reverts to the "_**sample**_" project. This is because both terminal sessions have the identical widget name, "**_sample_**". To address this issue effectively, need to delve deeper into the matter and explore alternative unique identifiers, aside from the **widget name**, to correctly determine the terminal session on which to set focus.

vaisakhkannan:
<img width="1291" alt="Screenshot 2024-01-30 at 11 37 36 AM" src="https://github.com/OpenLiberty/liberty-tools-intellij/assets/133869023/381998ef-16c6-40bd-96c4-f5d0d4cc4edd">

I can see the unique _display name_ instead of comparing it only with the _widget name_. The value is located within libertyModule -> shellWidget -> myTerminalTitle, and further inside the listeners → content → display name. I am currently working on obtaining the unique value for comparison.

vaisakhkannan:
There is a PR (https://github.com/OpenLiberty/liberty-tools-intellij/pull/637) open for the common issue (https://github.com/OpenLiberty/liberty-tools-intellij/issues/198). Once the above PR is merged, we can close this issue.

vaisakhkannan:
The PR https://github.com/OpenLiberty/liberty-tools-intellij/pull/637 has been merged, and the problem described in this issue has also been addressed with the same PR. I am now closing this issue.

PR #637, opened 2024-02-01
Fixes #198 -- Add code to shift focus correctly for each terminal

### LINKED ISSUE #198: Terminal focus doesn't shift and terminal renaming doesn't work

vaisakhkannan:
Fixes #198 -->  Added code to make sure the focus shifts correctly for each terminal when we perform dev-actions, also tested that terminal renaming works properly.

We can close these dependent issues #505, #506, once the PR is merged.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2023 IBM Corporation.
+ * Copyright (c) 2020, 2024 IBM Corporation.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
@@ -32,13 +32,10 @@ import org.xml.sax.SAXException;
 import javax.xml.parsers.ParserConfigurationException;
 import java.io.File;
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
+import java.util.*;
 
 public class LibertyProjectUtil {
-    private static Logger LOGGER = Logger.getInstance(LibertyProjectUtil.class);;
+    private static Logger LOGGER = Logger.getInstance(LibertyProjectUtil.class);
 
     enum BuildFileFilter {
         ADDABLE {
@@ -155,20 +152,18 @@ public class LibertyProjectUtil {
         ShellTerminalWidget widget = getTerminalWidget(libertyModule, terminalView);
         TerminalToolWindowManager manager = TerminalToolWindowManager.getInstance(project);
         ToolWindow toolWindow = manager.getToolWindow();
-        String widgetName = libertyModule.getName();
 
         ContentManager contentManager = toolWindow.getContentManager();
         Content[] contents = contentManager.getContents();
 
         int index = 0;
         for (int i = 0; i < contents.length; i++) {
-            String displayName = contents[i].getDisplayName();
-            if (widgetName.equals(displayName)) {
+            if (contents[i].getPreferredFocusableComponent().equals(widget)) {
                 index = i;
                 break;
             }
         }
-        if (contents.length > 0) {
+        if (contents.length > 0 && index > 0) {
             Content terminalContent = contents[index];
             contentManager.setSelectedContent(terminalContent);
             terminalContent.getComponent().requestFocus();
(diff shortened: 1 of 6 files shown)


dessina-devasia (dismissed):
Looks good for me

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
```diff
@@ -149,6 +153,27 @@ public static ShellTerminalWidget getTerminalWidget(Project project, LibertyModu
...
+                    break;
+                }
+            }
+            if (contents.length > 0 && index > 0) {
+                Content terminalContent = contents[index];
+                contentManager.setSelectedContent(terminalContent);
+                terminalContent.getComponent().requestFocus();
+            }
+        }
+
```
mrglavas, line 176: It would be better to separate the set focus code to its own method in `LibertyProjectUtil.java` which can be called from `getTerminalWidget()` in `LibertyGeneralAction.java`. Can perhaps try using `widget.hasFocus()` to check if the terminal already has focus before calling `setFocus()`.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
```diff
@@ -149,6 +153,27 @@ public static ShellTerminalWidget getTerminalWidget(Project project, LibertyModu
...
+        TerminalToolWindowManager manager = TerminalToolWindowManager.getInstance(project);
+        ToolWindow toolWindow = manager.getToolWindow();
+
+        if (toolWindow != null) {
+            ContentManager contentManager = toolWindow.getContentManager();
+            Content[] contents = contentManager.getContents();
+
+            int index = 0;
+            for (int i = 0; i < contents.length; i++) {
+                if (contents[i].getPreferredFocusableComponent().equals(widget)) {
```
mrglavas, line 165: This may no longer apply once the code has been refactored, but consider if `widget == null`. There should be guard for this entire block so that we don't spend time searching for a component that won't be found.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
mrglavas: Should also rename the `getTerminalWidget()` method in `LibertyGeneralAction.java` to reflect the fact that it is setting focus in addition to retrieving the terminal widget.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
vaisakhkannan: @mrglavas , Now Its checking `widget == null` in the if condition

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
vaisakhkannan: `widget.hasFocus()` always returns false, even though the corresponding terminal widget is `visible` and focused.

mrglavas: I see that you moved the code into `setFocusToWidget()` but its still being called directly by `getTerminalWidget()`. This meant to be moved up to the method you renamed in `LibertyGeneralAction.java`. Did you try this out? Was there an issue with doing this?

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
mrglavas: Thanks. I see that.

mrglavas (changes requested):
See recent comments.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
vaisakhkannan: I made the changes as per your suggestions

mrglavas (approved):
Looks good. Thanks for addressing my review comments.

aparnamichael (approved):

PR #662, opened 2024-02-23
Increase timeout before shutting down language servers

### LINKED ISSUE #648: Diagnostics are not always published when a file containing an error is opened.
I have observed this issue with Liberty Tools for IntelliJ ([useLSP4IJ1201](https://github.com/OpenLiberty/liberty-tools-intellij/tree/useLSP4IJ1201) branch) and Quarkus Tools v1.30.1. Occasionally when I open a file containing an error, no diagnostics are published. This doesn't seem to happen often but I've been able to reproduce it. To reproduce this, I open and close one file (specifically the `ServiceLiveHealthCheck` class with the `@Liveness` annotation removed from a MicroProfile starter project) repeatedly, allowing the MicroProfile language server to shutdown completely before opening the file again. Eventually when I open the file, no diagnostics will be displayed. I'm guessing there must be a timing issue and something with the language server/client isn't initialized around the time diagnostics would have been collected and published.

Screenshot of IntelliJ with the Liberty Tools plug-in installed:
image

Screenshot of IntelliJ with the Quarkus Tools plug-in installed:
image

turkeylurkey:
The value is in seconds.

Fixes #648

### DIFF
--- a/src/main/resources/META-INF/lsp.xml
+++ b/src/main/resources/META-INF/lsp.xml
@@ -16,6 +16,7 @@
                 class="io.openliberty.tools.intellij.lsp4mp.lsp.MicroProfileServer"
                 clientImpl="io.openliberty.tools.intellij.lsp4mp.lsp.MicroProfileLanguageClient"
                 serverInterface="org.eclipse.lsp4mp.ls.api.MicroProfileLanguageServerAPI"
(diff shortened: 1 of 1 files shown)


mrglavas (approved):
Looks good. Thanks.

Issue #681, opened 2024-03-07
Jakarta EE completion throws an IllegalArgumentException when typing within a method.


mrglavas:
Within IntelliJ, I'm seeing this exception appearing frequently while typing in a method. It's trying to compute the cursor context but failing when it attempts to convert `JavaCursorContextKind` enum values from LSP4MP to the LSP4Jakarta equivalent. This is specifically an issue with conversion of `JavaCursorContextKind.NONE` which has a value of 2000. The `forValue()` method assumes that none of the values are greater than the number of enum values and throws the exception when 2000 is passed in. In my opinion this is a bug in both LSP4MP and LSP4Jakarta but can be easily worked around.

```
java.lang.IllegalArgumentException: Illegal enum value: 2000
	at org.eclipse.lsp4jakarta.commons.JavaCursorContextKind.forValue(JavaCursorContextKind.java:230)
	at io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.PropertiesManagerForJakarta.adapt(PropertiesManagerForJakarta.java:126)
	at io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.PropertiesManagerForJakarta.javaCursorContext(PropertiesManagerForJakarta.java:86)
	at io.openliberty.tools.intellij.lsp4jakarta.lsp.JakartaLanguageClient.lambda$getJavaCompletion$3(JakartaLanguageClient.java:90)
	at org.microshed.lsp4ij.internal.PromiseToCompletableFuture.lambda$nonBlockingReadActionPromise$2(PromiseToCompletableFuture.java:137)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$OTelMonitor.callWrapped(NonBlockingReadActionImpl.java:840)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$OTelMonitor$MonitoredComputation.call(NonBlockingReadActionImpl.java:872)
	at com.intellij.openapi.application.impl.NonBlockingReadActionImpl$Submission.insideReadAction(NonBlockingReadActionImpl.java:604)
	... 27 more
```

Current implementation of `JavaCursorContextKind.forValue()` in LSP4Jakarta:

```
    public static JavaCursorContextKind forValue(int value) {
        JavaCursorContextKind[] allValues = JavaCursorContextKind.values();
        if (value < 1 || value > allValues.length)
            throw new IllegalArgumentException("Illegal enum value: " + value);
[...]

PR #682, opened 2024-03-08
Implementing a workaround for an issue with JavaCursorContextKind.forValue().

mrglavas:
Resolves https://github.com/OpenLiberty/liberty-tools-intellij/issues/681. This is achieved by treating `JavaCursorContextKind.NONE` as a special case, avoiding the issue with retrieving the enum value through the `forValue()` method.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/PropertiesManagerForJakarta.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/PropertiesManagerForJakarta.java
@@ -122,8 +122,19 @@ public final class PropertiesManagerForJakarta {
 
     private JavaCursorContextResult adapt(org.eclipse.lsp4mp.commons.JavaCursorContextResult contextResult) {
         if (contextResult != null) {
-            var kind = contextResult.getKind();
-            return new JavaCursorContextResult(JavaCursorContextKind.forValue(kind.getValue()), contextResult.getPrefix());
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/PropertiesManagerForJakarta.java
```diff
@@ -122,8 +122,19 @@ private MicroProfileJavaCompletionParams adapt(JakartaJavaCompletionParams param
 
     private JavaCursorContextResult adapt(org.eclipse.lsp4mp.commons.JavaCursorContextResult contextResult) {
         if (contextResult != null) {
-            var kind = contextResult.getKind();
-            return new JavaCursorContextResult(JavaCursorContextKind.forValue(kind.getValue()), contextResult.getPrefix());
+            return new JavaCursorContextResult(adapt(contextResult.getKind()), contextResult.getPrefix());
```
TrevCraw, line 125: Since the new `adapt()` method can return `null`, do we know if `null` is safe to pass as the `JavaCursorContextKind` parameter for the `JavaCursorContextResult` constructor?

mrglavas: It is safe. It's just writing the value to a field and nothing else. The default constructor passes null values to that constructor already.

turkeylurkey (approved):

TrevCraw (approved):

PR #709, opened 2024-04-04
Fixed to throw IllegalStateException

### LINKED ISSUE #707: The validateTestReportExists method should throw an Exception if no test report is found
Currently, the validateTestReportExists method (https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java#L377) does not trigger a test failure if the test report is not found.

Suggestion to throw the following exception (pulled from LT Eclipse) after the loop fails to find the test report:
```
throw new IllegalStateException("Timed out waiting for test report: " + pathToTestReport + " file to be created.");
```

As part of this issue, we should confirm that each of the tests relying on this method fail when a test report is not created.

vaisakhkannan:
Fixes #707 
I confirmed that each of the tests relying on the method **validateTestReportExists** (https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java#L377)  failed when a test report was not created.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
@@ -388,13 +388,12 @@ public class TestUtils {
                     Thread.sleep(retryIntervalSecs * 1000);
                 } catch (Exception e) {
                     e.printStackTrace(System.out);
-                    continue;
                 }
-                continue;
+            } else {
+                return;
             }
-
-            return;
         }
+        throw new IllegalStateException("Timed out waiting for test report: " + pathToTestReport + " file to be created.");
     }
 
     /**


dessina-devasia (approved):
Looks good for me

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -388,13 +388,12 @@ public static void validateTestReportExists(Path pathToTestReport) {
...
-                    continue;
                 }
-                continue;
+            } else {
+                return;
             }
-
-            return;
         }
+        throw new IllegalStateException("Timed out waiting for test report: " + pathToTestReport + " file to be created.");
```
dessina-devasia, line 396: how do we further handled the exception

dessina-devasia (changes requested):

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
vaisakhkannan: When I investigated, I found that it's not always necessary to handle exceptions using a **_try-catch_** block for unchecked exceptions like **_IllegalStateException_**.


In this case, we don't need to **catch** the exception because we've checked the condition 100 times (and the **fileExists** value remains `false` throughout). Therefore, we can `throw` the exception directly without implementing further catch mechanisms for the corresponding test, resulting in **test failure**.

dessina-devasia (approved):

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
mrglavas, line 396: Thanks for the demo this morning. This looks correct, but will wait for the go-ahead from @TrevCraw that `main` is open again for PR merges before approving.

vaisakhkannan: Sure @mrglavas , we can wait. Thank You

mrglavas: `main` is open again, so I'm planning to approve now.

mrglavas (approved):
Looks good. Thanks.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
vaisakhkannan: Thanks @mrglavas

PR #714, opened 2024-04-09
Update Gradle IntelliJ Plugin version number

### LINKED ISSUE #668: Update Gradle IntelliJ Plugin version number to use the latest
We currently specify `id 'org.jetbrains.intellij' version '1.13.3'` in our build.gradle. When we start IntelliJ we see a message similar to the following: 
```
> Task :initializeIntelliJPlugin
[gradle-intellij-plugin :initializeIntelliJPlugin] Gradle IntelliJ Plugin is outdated: 1.13.3. Update `org.jetbrains.intellij` to: 1.17.2
```
We should update to the latest version of Gradle IntelliJ Plugin or document in build.gradle why we stick with a specific version.

anusreelakshmi934:
Fixes #668


anusreelakshmi934:
Updated the Gradle IntelliJ Plugin version to `1.17.3` which is the latest in the build.gradle file. Was not able to run the extension after the change, so tried changing the `distributionUrl` to `https\://services.gradle.org/distributions/gradle-7.6-all.zip` and the extesnion ran. The PR builds are failing. 
But When running locally all the tests are getting success but encountering a NoSuchMethodException.

anusreelakshmi934:
changed the distributionUrl to the latest gradle version` distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-all.zip` . Also Changed the compileTestKotlin in build.gradle to tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all . it ensures that the JVM target is set to 17 for all Kotlin compilation tasks, including both main and test sources. while compileTestKotlin only sets the JVM target to 17 for Kotlin code in the test sources.

build.gradle
```diff
@@ -55,7 +55,7 @@ configurations {
         exclude group: 'org.slf4j', module: 'slf4j-api'
     }
 
-    compileTestKotlin {
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
```
aparnamichael, line 58: Can you please explain the reason to change this compileTestKotlin to tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all

What is the difference it make by changing this.

anusreelakshmi934: tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all ensures that the JVM target is set to 17 for all Kotlin compilation tasks, including both main and test sources. while compileTestKotlin only sets the JVM target to 17 for Kotlin code in the test sources.

aparnamichael:
Looks good to me.

aparnamichael (dismissed):

build.gradle
TrevCraw: @aparnamichael Thank you for asking this question, I was wondering about this myself.

@anusreelakshmi934 For clarification, is the error you mention here the same as the `NoSuchMethodException` mentioned in your earlier comments?

 Also, when looking over the StackOverflow article provided, I noticed it was focused on Android projects and the answer had this line:
 "_There's no compileKotlin task, which is usually created for the main source set in ordinary Java + Kotlin projects._"
 I believe our project would be an "ordinary Java + Kotlin" project. Did you try using `compileKotlin` on this line?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.7-all.zip
```
TrevCraw, line 3: what is the difference between specifying `bin` and specifying `all`?

Wondering if it would be better to go with `gradle-8.7-bin.zip`

TrevCraw (commented):
You mentioned this in one of your earlier comments:

> Also we are getting > Task :initializeIntelliJPlugin SKIPPED in the log

What was the resolution for this? Has this gone away? Or can we ignore it, and if so, why?

build.gradle
anusreelakshmi934: Initially, I changed the version of `'org.jetbrains.intellij'` to `'1.17.3'`, which led to an error stating "Unsupported class file major version 63". Investigating upon that, I found a solution on [this GitHub thread](https://github.com/react-native-community/cli/issues/1719), suggesting to modify the `distributionUrl` in the 'gradle-wrapper.properties' file. I opted for `gradle-7.6-all.zip`, but this change resulted in a `NoSuchMethodException.`

Further exploring, I updated the version to the latest `gradle-8.7-all.zip` . Upon changing that i got an error - `Could not find method kotlinOptions() for arguments [build_7jqpmvy23lx9badxvjae0ix41$_run_closure2$_closure22$_closure23@13223cc3] on configuration ':compileTestKotlin' of type org.gradle.api.internal.artifacts.configurations.DefaultUnlockedConfiguration`. To resolve this, I replaced `compileTestKotlin` with `tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all.` Now everything works fine without any exceptions or errors. 
I also attempted using compileKotlin, but faced the same aforementioned error.

gradle/wrapper/gradle-wrapper.properties
anusreelakshmi934: The `bin` distribution contains only the binaries necessary to run Gradle, while the `all` distribution includes additional files such as documentation, source code, and samples. 
Yes I tried with `gradle-8.7-bin.zip` it works fine. so we can use that.

anusreelakshmi934:
> You mentioned this in one of your earlier comments:
> 
> > Also we are getting > Task :initializeIntelliJPlugin SKIPPED in the log
> 
> What was the resolution for this? Has this gone away? Or can we ignore it, and if so, why?

When I checked the PR builds of this on GitHub, I noticed that the task `:initializeIntelliJPlugin` is not being skipped . Even in my local environment, I've observed instances where it isn't skipped always. I conducted some investigation but couldn't find any relevant information. However, I came across other links discussing some other issues, such as [this one](https://github.com/JetBrains/intellij-platform-gradle-plugin/issues/1515) where `Task :initializeIntelliJPlugin` is mentioned as being skipped in the relevant log output or stack trace, and [this one](https://youtrack.jetbrains.com/issue/WEB-65129) where it's noted in the build output that the task is skipped. In these discussions, nobody explicitly mentions `Task :initializeIntelliJPlugin` being skipped as an error or issue. Based on this information, I suppose this might be something we can overlook.

build.gradle
TrevCraw: @anusreelakshmi934 thanks for the explanation. Please add a comment explaining what this line is doing.

anusreelakshmi934: Added the explanation

TrevCraw:
Hi @anusreelakshmi934, I was reading a bit more about updating the Gradle wrapper version. This documentation is particularly useful: https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper.

I think we should follow the recommended way of updating the gradle wrapper - "run the wrapper task and provide the target Gradle version" (note, it looks like the `wrapper` task has to be run twice to completely update). Doing this will update all the other associated gradle wrapper files (gradle-wrapper.jar, gradlew, gradlew.bat).

anusreelakshmi934:
> run the wrapper task

Hi @TrevCraw . I have updated gradle wrapper using `gradle wrapper`. Thereby updating the associated files (gradlew.bat, gradlew, gradle-wrapper.properties, gradle-wrapper.jar) automatically.

TrevCraw (approved):

aparnamichael (approved):
Looks good to me

anusreelakshmi934:
@vaisakhkannan

vaisakhkannan:
@anusreelakshmi934

anusreelakshmi:
@anusreelakshmi934

PR #744, opened 2024-05-09
Refactor the UI test method pasteOnActiveWindow

turkeylurkey:
There is a fix which is needed in multiple locations. This effort is to move the code to one common location to assist future support.


mrglavas (approved):
Looks good. Thanks.

PR #745, opened 2024-05-10
Merge the 2024.1.1 support into main branch

turkeylurkey:


turkeylurkey:
The movie associated with this failure shows IntelliJ running quite slowly and jerky. The test seems to pass so it is a mystery why this failure appears. Possible timing problem.
image

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/completion/IJavaCompletionParticipant.java
```diff
@@ -24,6 +24,7 @@
  * @author datho7561
  */
 public interface IJavaCompletionParticipant {
+	public static final ExtensionPointName<IJavaCompletionParticipant> EP_NAME = ExtensionPointName.create("open-liberty.intellij.javaCompletionParticipant");
```
mrglavas, line 27: This extension point is unused. The real one is in `JavaCompletionDefinition`. I know this keeps coming back every time we synch from Quarkus Tools. We deleted it the last time. It's not harmful but it is misleading (potentially confusing) to someone reading the code.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/diagnostics/IJavaDiagnosticsParticipant.java
```diff
@@ -24,6 +25,8 @@
  */
 public interface IJavaDiagnosticsParticipant {
 
+	public static final ExtensionPointName<IJavaDiagnosticsParticipant> EP_NAME = ExtensionPointName.create("open-liberty.intellij.javaDiagnosticsParticipant");
```
mrglavas, line 28: This is unused. The real extension point is in `JavaDiagnosticsDefinition` which implements `IJavaDiagnosticsParticipant.

mrglavas (approved):
Looks good. There are a few unused extension points that we might want to clean-up but we can do that later.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/completion/IJavaCompletionParticipant.java
turkeylurkey: Ok, we should mark this and the other one as unused so this doesn't happen again.

mrglavas: I'm not sure what you mean by marking it, but I guess this could be something we include in our process for synching up lsp4mp4ij. It's a change we made in LTI to support a common code path for both MP and Jakarta EE completion.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/diagnostics/IJavaDiagnosticsParticipant.java
turkeylurkey: Added comment for the benefit of future maintainers.

mrglavas (approved):
Looks good with the updates.

TrevCraw (approved):

Issue #635, opened 2024-01-31
Apply catch/rethrow IndexNotReadyException, ProcessCanceledException and CancellationException to LSP4Jakarta code action unification branch.


mrglavas:
Review the code that has been merged into the https://github.com/OpenLiberty/liberty-tools-intellij/tree/code-action-unification-2 branch and apply the catch/rethrow of IndexNotReadyException, ProcessCanceledException and CancellationException to all the locations that still require it. So that this work can be completed in one pass we should wait and only start working on this work item once development has been completed on the branch. See parent issue: https://github.com/OpenLiberty/liberty-tools-intellij/issues/627 for more details.

mrglavas:
Resolved by https://github.com/OpenLiberty/liberty-tools-intellij/pull/750.

PR #750, opened 2024-05-14
Jakarta EE quick fixes: Adding a catch and rethrow for exceptions that should be propagated up the call stack.

mrglavas:
Added a catch/rethrow of IndexNotReadyException, ProcessCanceledException, CancellationException to `resolveCodeAction()` and related methods. Resolves https://github.com/OpenLiberty/liberty-tools-intellij/issues/635.

Also cleaned up imports and updated copyrights.


turkeylurkey (approved):
That is code.

TrevCraw (approved):
As discussed earlier, the duplicate code is OK for now, but we should look to refactor the common code for `resolveCodeAction()` in the future.

mrglavas:
> As discussed earlier, the duplicate code is OK for now, but we should look to refactor the common code for `resolveCodeAction()` in the future.

Agreed an we should look at this more holistically than this one example. I opened: https://github.com/OpenLiberty/liberty-tools-intellij/issues/754 to track this activity.

PR #771, opened 2024-05-23
Wait for list of options to be populated.

### LINKED ISSUE #768: New Project bug in automated testing of Add Project using search.
When using shift-shift menu to add a project the robot searches for the task `Liberty: Add project to the tool window` then selects IntelliJ New Project.
image
Log failure
```
org.opentest4j.AssertionFailedError: testRunTestsActionUsingSearch:validateProjectStarted: Timed out while waiting for project with resource URI api/resourceand port 9080 to become available.
```
Parent #764

### LINKED ISSUE #769: Server has stopped when trying to run tests in automated build
The action to run tests using the shift-shift menu tries to start the server which it does but when it then tries to run the tests the robot finds the server is not running because it shutdown three seconds after it started. Then the clean-up code tries to stop the server and correctly reports the server is not running.
image

Log message for failed test
```
org.opentest4j.AssertionFailedError: testRunTestsActionUsingSearch:validateProjectStarted: Timed out while waiting for project with resource URI api/resourceand port 9080 to become available.
```
Parent #764

turkeylurkey:
Three green builds in a row. Hopefully not due to luck.

image

Just to explain the fix, when the robot pastes the search text into the search box in the following screen the robot pauses 1s before reading and searching the options underneath. Likely the options are generated asynchronously on a separate thread. 
image

Fixes #768
Fixes #769

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
@@ -129,8 +129,8 @@ public abstract class SingleModJakartaLSTestCommon {
     /**
      * Tests Jakarta Language Server quick fix support in a Java source file
      */
-    @Test
-    @Video
+//    @Test
+//    @Video
     public void testJakartaQuickFixInJavaPart() {
         String publicString = "public Response getProperties() {";
         String privateString = "private Response getProperties() {";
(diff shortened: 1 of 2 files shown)


src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
```diff
@@ -129,8 +129,8 @@ public void testJakartaDiagnosticsInJavaPart() {
     /**
      * Tests Jakarta Language Server quick fix support in a Java source file
      */
-    @Test
-    @Video
+//    @Test
+//    @Video
```
mrglavas, line 133: If you haven't already done so, we should have an issue open to restore this test so we don't lose track that we disabled it here.

mrglavas (approved):
Looks good. Thanks.

src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
turkeylurkey: I was thinking we should reopen the issue that Trevor closed yesterday and add the note to re-add this test.

TrevCraw (approved):

turkeylurkey:
Three green linux runs.

PR #772, opened 2024-05-23
Wait 60s when starting server in the test Run Tests and wait 1s for search for Liberty:Start

### LINKED ISSUE #769: Server has stopped when trying to run tests in automated build

turkeylurkey:
Three "green" runs in a row hopefully demonstrates this fix works. The first run had the quick fix error so technicaslly not green but it did not have tis error. 

image

This fix works the same as that for #768 

Fixes #769

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -1596,6 +1596,7 @@ public class UIBotTestUtils {
                 JTextFieldFixture searchField = projectFrame.textField(JTextFieldFixture.Companion.byType(), Duration.ofSeconds(10));
                 searchField.click();
                 searchField.setText(action);
+                TestUtils.sleepAndIgnoreException(1); // allow the search for 'action' to complete
 
                 // Wait for the desired action to show in the search output frame and click on it.
                 RepeatUtilsKt.waitFor(Duration.ofSeconds(20),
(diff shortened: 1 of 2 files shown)


mrglavas (approved):
Looks good. Thanks.

src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
```diff
@@ -1596,6 +1596,7 @@ public static void runActionFromSearchEverywherePanel(RemoteRobot remoteRobot, S
                 JTextFieldFixture searchField = projectFrame.textField(JTextFieldFixture.Companion.byType(), Duration.ofSeconds(10));
                 searchField.click();
                 searchField.setText(action);
+                TestUtils.sleepAndIgnoreException(1); // allow the search for 'action' to complete
```
TrevCraw, line 1599: This is the exact same line changed in https://github.com/OpenLiberty/liberty-tools-intellij/pull/771. This will create a merge conflict since you have a different comment here.

turkeylurkey: So it is. The change in SingleModMPProjectTestCommon.java is not needed so I will cancel this pull request.

PR #799, opened 2024-06-05
Simplify the API for custom params in LibertyModule

### LINKED ISSUE #798: Improve the API of the Liberty Module custom parameters facility
The LibertyModule class holds a string for the custom start parameters. There is a getter and a setter but there is also a method to clear the string. This makes it harder to find locations that set the custom start parameters because there are multiple ways to change them. We should remove `clearCustomStartParams()` and just use `setCustomStartParams()` instead.

turkeylurkey:
Fixes #798


TrevCraw (approved):

mrglavas (approved):
Looks good. Thanks.

PR #828, opened 2024-06-17
Fixed using Deprecated Maven Methods which are Marked for Removal

### LINKED ISSUE #823: Stop Using Deprecated Maven Methods which are Marked for Removal
The **Liberty-Tools-Intellij** build step uses two deprecated maven methods that are marked for removal. We need to fix this issue so that we won’t receive warnings in the future.

The error in the logs is

```
> Task :compileKotlin NO-SOURCE
/home/runner/work/liberty-tools-intellij/liberty-tools-intellij/liberty-tools-intellij/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java:207: warning: [removal] getMavenHome() in MavenGeneralSettings has been deprecated and marked for removal
        String mavenHome = mavenSettings.getMavenHome();
                                        ^
/home/runner/work/liberty-tools-intellij/liberty-tools-intellij/liberty-tools-intellij/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java:250: warning: [removal] resolveMavenHomeDirectory(String) in MavenUtil has been deprecated and marked for removal

> Task :compileJava
        File mavenHomeFile = MavenUtil.resolveMavenHomeDirectory(customMavenHome); // when customMavenHome path is invalid it returns null
                                      ^
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
Note: Some input files use unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
2 warnings
```

vaisakhkannan:
Fixes #823 
Updated `getMavenHome()` method to `getMavenHomeType()` method and
Updated `resolveMavenHomeDirectory()` method to `getMavenHomeFile()` method

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
@@ -209,13 +209,13 @@ public class LibertyMavenUtil {
     final static String wrappedMaven = "Use Maven wrapper";
     public static String getMavenSettingsCmd(Project project, VirtualFile buildFile) throws LibertyException {
         MavenGeneralSettings mavenSettings = MavenWorkspaceSettingsComponent.getInstance(project).getSettings().getGeneralSettings();
-        @NotNull MavenHomeType mavenHome = mavenSettings.getMavenHomeType();
-        if (wrappedMaven.equals(mavenHome.getTitle())) {
+        @NotNull MavenHomeType mavenHomeType = mavenSettings.getMavenHomeType();
+        if (wrappedMaven.equals(mavenHomeType.getTitle())) {
             // it is set to use the wrapper
             return getLocalMavenWrapper(buildFile);
         } else {
             // try to use maven home path defined in the settings
-            return getCustomMavenPath(project, mavenHome.getTitle());
+            return getCustomMavenPath(project, mavenHomeType.getTitle());
         }
     }
 
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
@@ -252,7 +252,8 @@ public class LibertyMavenUtil {
      * @throws LibertyException
      */
     private static String getCustomMavenPath(Project project, @NotNull MavenHomeType customMavenHome) throws LibertyException {
-        StaticResolvedMavenHomeType resolvedMavenHomeType = getResolvedMavenHomeType(customMavenHome);
+        // If 'customMavenHome' is not an instance of 'StaticResolvedMavenHomeType',it using a fallback method (staticOrBundled) to return default 'BundledMaven3.INSTANCE'.
+        StaticResolvedMavenHomeType resolvedMavenHomeType = staticOrBundled(customMavenHome);
         File mavenHomeFile = MavenUtil.getMavenHomeFile(resolvedMavenHomeType);
         // when customMavenHome path is invalid it returns null
         if (mavenHomeFile == null) {
@@ -294,20 +295,6 @@ public class LibertyMavenUtil {
             throw new LibertyException(String.format("Could not execute the Maven executable %s. Make sure a valid path is configured " +
                     "inside IntelliJ Maven preferences.", mavenExecutable.getAbsolutePath()), translatedMessage);
         }
-    }
-
-    /**
-     * This method checks if the provided 'MavenHomeType' is an instance of 'StaticResolvedMavenHomeType'.
-     * If it is, it returns the casted object. Otherwise, it uses a fallback method (staticOrBundled) to return default 'BundledMaven3.INSTANCE'
-     * @param customMavenHome the custom MavenHomeType to be resolved
-     * @return the resolved StaticResolvedMavenHomeType
-     */
-    private static StaticResolvedMavenHomeType getResolvedMavenHomeType(MavenHomeType customMavenHome) {
-        if (customMavenHome instanceof StaticResolvedMavenHomeType) {
-            return (StaticResolvedMavenHomeType) customMavenHome;
-        }
-        // If 'customMavenHome' is not an instance of 'StaticResolvedMavenHomeType', return default 'BundledMaven3.INSTANCE'.
-        return staticOrBundled(customMavenHome);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
```diff
@@ -204,13 +207,13 @@ private static boolean containerVersion(String version){
     final static String wrappedMaven = "Use Maven wrapper";
     public static String getMavenSettingsCmd(Project project, VirtualFile buildFile) throws LibertyException {
         MavenGeneralSettings mavenSettings = MavenWorkspaceSettingsComponent.getInstance(project).getSettings().getGeneralSettings();
-        String mavenHome = mavenSettings.getMavenHome();
-        if (wrappedMaven.equals(mavenHome)) {
+        @NotNull MavenHomeType mavenHome = mavenSettings.getMavenHomeType();
```
aparnamichael, line 210: Rename this variable to mavenHomeType

src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
```diff
@@ -246,8 +249,8 @@ private static String getLocalMavenWrapper(VirtualFile buildFile) throws Liberty
      * @return Maven path to be executed or an exception to display
      * @throws LibertyException
      */
-    private static String getCustomMavenPath(Project project, String customMavenHome) throws LibertyException {
-        File mavenHomeFile = MavenUtil.resolveMavenHomeDirectory(customMavenHome); // when customMavenHome path is invalid it returns null
+    private static String getCustomMavenPath(Project project, StaticResolvedMavenHomeType customMavenHome) throws LibertyException {
+        File mavenHomeFile = MavenUtil.getMavenHomeFile(customMavenHome); // when customMavenHome path is invalid it returns null
```
aparnamichael, line 253: Instead of changing parameter type to StaticResolvedMavenHomeType, can we revert it back to string and is it possible to call "MavenUtil.getMavenHomeFile" method in a similar way written inside the "MavenUtil.resolveMavenHomeDirectory" deprecated method.

@Nullable
@Deprecated(forRemoval = true)
public static File resolveMavenHomeDirectory(@Nullable String overrideMavenHome) {
  if (!isEmptyOrSpaces(overrideMavenHome)) {
    //noinspection HardCodedStringLiteral
    return MavenUtil.getMavenHomeFile(staticOrBundled(resolveMavenHomeType(overrideMavenHome)));
  }

vaisakhkannan:
@aparnamichael , When I compared the implementation of the old deprecated method `(resolveMavenHomeDirectory)`, I noticed it uses a `staticOrBundled` method. What this method actually does is..

1. It attempts to cast the current object (this) to `StaticResolvedMavenHomeType`.
2. If the cast is successful, it returns the casted object.
3. If the cast fails (i.e., this is not an instance of `StaticResolvedMavenHomeType`), it returns '`BundledMaven3`'.
Screenshot 2024-06-20 at 1 56 23 PM


Using the same approach, I modified the code to return the default 'BundledMaven3.INSTANCE'

If we use `null` as the default return value, we will get an `IllegalArgumentException` if the casting fails in the future, as shown in the image. I encountered this issue during testing because I commented out the relevant code below.

https://github.com/OpenLiberty/liberty-tools-intellij/blob/54d1e896f432156a058642183e48b222a86b44c9/src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java#L208C9-L211C10

Screenshot 2024-06-20 at 2 24 48 PM



I believe this solution is better now. Please review my changes.

aparnamichael (approved):

mrglavas:
> I believe this solution is better now. Please review my changes.

Thanks for digging deeper to align the behaviour of the new API usage with the old one.

mrglavas (approved):
Looks good now. Thanks!

src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
```diff
@@ -289,6 +296,20 @@ private static String getCustomMavenPath(Project project, String customMavenHome
...
+     * If it is, it returns the casted object. Otherwise, it uses a fallback method (staticOrBundled) to return default 'BundledMaven3.INSTANCE'
+     * @param customMavenHome the custom MavenHomeType to be resolved
+     * @return the resolved StaticResolvedMavenHomeType
+     */
+    private static StaticResolvedMavenHomeType getResolvedMavenHomeType(MavenHomeType customMavenHome) {
+        if (customMavenHome instanceof StaticResolvedMavenHomeType) {
+            return (StaticResolvedMavenHomeType) customMavenHome;
+        }
+        // If 'customMavenHome' is not an instance of 'StaticResolvedMavenHomeType', return default 'BundledMaven3.INSTANCE'.
+        return staticOrBundled(customMavenHome);
```
TrevCraw, line 310: I have a couple questions around using the `staticOrBundled()` method here.
1. Is this method intended to be part of the public API for the jetbrains maven library? Or is it an internal method?
2. If the behaviour of `staticOrBundled()` is to attempt to cast the input param to `StaticResolvedMavenHomeType` or return `BundledMaven3.INSTANCE`, could we not just do that ourselves here? We already check for the instance type with the `if` statement. If that is `false`, I think that means that we can assume `BundledMaven3.INSTANCE` will be returned from `staticOrBundled()`... or am I missing something?

To summarize, if the `staticOrBundled()` method is an internal method of the jetbrains maven library, we probably should not use it and just mock its behaviour in the `getResolvedMavenHomeType()` method. If it is intended to be part of the publicly available API, then I think we can just use the `staticOrBundled()` method directly from `getCustomMavenPath()` and cut out the "middle man" method `getResolvedMavenHomeType()`.

vaisakhkannan: 1. Yes @TrevCraw , `staticOrBundled()` method intended to be part of the public API for the jetbrains maven library and not an internal method
Screenshot 2024-06-25 at 1 54 59 PM

2. I modified the code to use the `staticOrBundled()` method directly from `getCustomMavenPath(`) and removed  method`(getResolvedMavenHomeType())` that I created.

TrevCraw: Hi @vaisakhkannan , the changes look good. Did you re-test with these new changes?

vaisakhkannan: Yes @TrevCraw , I did testing in all 3 maven settings and its working fine

aparnamichael (approved):
Looks good to me.

mrglavas (approved):
Looks good. Thanks.

TrevCraw (approved):

PR #833, opened 2024-06-20
Parameterization for Running the LTI Build Fixes #783

### LINKED ISSUE #783: Determine Parameterization for Running the LTI Build
Determine how to run the LTI build by using 

- Lsp4ij plugin from marketplace 
- Lsp4ij plugin build locally

anusreelakshmi934:
Fixes #783 - Parameterize Issue

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -2,6 +2,11 @@ name: Build
 
 on:
   workflow_dispatch:
+    inputs:
+      useLocalPlugin:
+        description: 'Use lsp4ij locally'
+        required: true
+        default: 'false'
   push:
     branches: '**'
   pull_request:
@@ -24,6 +29,8 @@ jobs:
           - runtime: windows
             os: windows-latest
             reportName: windows-test-report
+    env:
+      USE_LOCAL_PLUGIN: ${{ github.event.inputs.useLocalPlugin || 'false' }}
     steps:
       - name: Configure pagefile
         if: contains(matrix.os, 'windows')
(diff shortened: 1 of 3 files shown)


.github/workflows/build.yaml
```diff
@@ -39,9 +46,28 @@ jobs:
...
+          ref: main
+      - name: 'Build Lsp4ij'
+        if: env.USE_LOCAL_PLUGIN == 'true'
+        working-directory: ./lsp4ij
+        run: bash ./gradlew buildPlugin
+      - name: 'Unzip lsp4ij file'
+        if: env.USE_LOCAL_PLUGIN == 'true'
+        working-directory: ./lsp4ij/build/distributions
+        run: |
+          unzip -o '*.zip' -d .
```
mrglavas, line 66: I know I might be getting confused by the number of prototypes we've had for these workflows but I thought our continuous integration builds would support building from PRs as well as the current `main` branch. I'm not seeing that logic here. Do we have this somewhere else (on another branch or PR)? Can it all be integrated into the same workflow?

build.gradle
```diff
@@ -140,7 +140,10 @@ intellij {
     // For a full list of IntelliJ IDEA releases please see https://www.jetbrains.com/intellij-repository/releases
     version = '2024.1.3'
 
-    plugins = ['java', 'maven', 'gradle-java', 'properties', 'terminal', 'org.jetbrains.idea.maven', 'com.intellij.gradle', 'com.redhat.devtools.lsp4ij:' + lsp4ijVersion + '@nightly']
+    def lsp4ij = project.hasProperty('useLocal') && project.property('useLocal') == 'true' ?
+            file("../lsp4ij/build/distributions/LSP4IJ/") :
+            'com.redhat.devtools.lsp4ij:' + lsp4ijVersion + '@nightly'
```
mrglavas, line 145: I know you didn't change this in this PR, but noticed this is hard-coded here to pull the LSP4iJ version from the `Nightly` channel. When we release we definitely want to be depending on a `Stable` release of LSP4iJ for our build and testing. I wonder if we could separate that in `build.gradle` like the `lsp4ijVersion` variable that's being read here.

.github/workflows/build.yaml
anusreelakshmi934: > I know I might be getting confused by the number of prototypes we've had for these workflows but I thought our continuous integration builds would support building from PRs as well as the current `main` branch. I'm not seeing that logic here. Do we have this somewhere else (on another branch or PR)? Can it all be integrated into the same workflow and controlled by a build parameter?

This ticket was primarily for determining the parameterization required to run the LTI builds. For easy prototyping, I tested with checking out the main branch only. There is a separate ticket for running the LTI tests against each open lsp4ij PR.

build.gradle
anusreelakshmi934: > I know you didn't change this in this PR, but noticed this is hard-coded here to pull the LSP4iJ version from the `Nightly` channel. When we release we definitely want to be depending on a `Stable` release of LSP4iJ for our build and testing. I wonder if we could separate that in `build.gradle` like the `lsp4ijVersion` variable that's being read here.

Since we have not yet decided which version of Lsp4ij to use for our release, we are currently specifying the nightly version. Once we decide to proceed with the stable release, we can remove the `@nightly` tag and specify the stable version. For now, we are using `@nightly` to test our functionalities. When we move forward with our release, we will switch to the stable version.

TrevCraw: @anusreelakshmi934 I think @mrglavas 's suggestion is to parameterize the build channel the same way the `lsp4ijVersion` is parameterized  - i.e. the values would be `@nightly` or `@stable`.

mrglavas: @TrevCraw That is exactly what I was suggesting.

anusreelakshmi934: Yes. I have updated the `lsp4ijVersion` variable at the top of build.gradle by appending `@nightly` along with the version number. If we are to use the stable version we can remove that `@nightly` and just specify the version alone.

anusreelakshmi934:
Added all these changes to a Umbrella PR - https://github.com/OpenLiberty/liberty-tools-intellij/pull/840. Hence closing this PR

PR #834, opened 2024-06-20
Set up a GHA workflow as a cron job - Fixes #756

### LINKED ISSUE #756: Set up a GHA workflow as a cron job
Automate GHA workflows for daily cron jobs, facilitating the initiation of lsp4ij integration builds.

anusreelakshmi934:
Fixes #756 - Set up cron job

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,7 +1,19 @@
 name: Build
 
 on:
+  workflow_call:
+    inputs:
+      useLocalPlugin:
+        description: 'Use lsp4ij locally'
+        required: true
+        type: boolean
+        default: false
   workflow_dispatch:
+    inputs:
+      useLocalPlugin:
+        description: 'Use lsp4ij locally'
+        required: true
+        default: 'false'
   push:
     branches: '**'
   pull_request:
@@ -24,6 +36,8 @@ jobs:
           - runtime: windows
             os: windows-latest
             reportName: windows-test-report
+    env:
+      USE_LOCAL_PLUGIN: ${{ inputs.useLocalPlugin || 'false' }}
     steps:
       - name: Configure pagefile
         if: contains(matrix.os, 'windows')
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -1,7 +1,19 @@
...
+        description: 'Use lsp4ij locally'
+        required: true
+        type: boolean
+        default: false
   workflow_dispatch:
+    inputs:
+      useLocalPlugin:
+        description: 'Use lsp4ij locally'
+        required: true
+        default: 'false'
```
mrglavas, line 16: Is there a reason this wasn't declared as a boolean like the input in the `workflow_call`?

.github/workflows/build.yaml
```diff
@@ -24,6 +36,8 @@ jobs:
           - runtime: windows
             os: windows-latest
             reportName: windows-test-report
+    env:
+      USE_LOCAL_PLUGIN: ${{ inputs.useLocalPlugin || 'false' }}
```
mrglavas, line 40: Wondering if there's something I'm missing here. Normally a boolean expression of the form `X || false` is always equal to `X`. But then I noticed `'false'` is quoted like a string and in some scripting languages using a non-null string in a boolean expression is equal to `true`.

mrglavas: See definition of literals here: https://docs.github.com/en/actions/learn-github-actions/expressions#literals and the statement: 

> Note that in conditionals, falsy values (false, 0, -0, "", '', null) are coerced to false and truthy (true and other non-falsy values) are coerced to true.

It would seem that `'false'` in a conditional statement would be coerced to `true`.

.github/workflows/cronJob.yaml
```diff
@@ -0,0 +1,11 @@
+name: Cron Job
+
+on:
+  schedule:
+    - cron: '0 0 * * *'
```
mrglavas, line 5: I had to search around to learn that this means run every midnight. I wonder if we could add a comment which states that or whatever the current schedule is. Might even be good to have a link back to the docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule so that someone who wants to change the schedule in the future can better understand the syntax.

.github/workflows/build.yaml
anusreelakshmi934: Initially, I had defined the workflow dispatch event as a string in the initial commits. However, it can also be declared as a boolean. I have now changed it to a boolean.

.github/workflows/build.yaml
anusreelakshmi934: Got it. I have changed it to boolean.

.github/workflows/cronJob.yaml
anusreelakshmi934: > I had to search around to learn that this means run every midnight. I wonder if we could add a comment which states that or whatever the current schedule is. Might even be good to have a link back to the docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onschedule so that someone who wants to change the schedule in the future can better understand the syntax.

Sure. Added the comment and the reference link

anusreelakshmi934:
Added all these changes to a Umbrella PR - https://github.com/OpenLiberty/liberty-tools-intellij/pull/840. Hence closing this PR

Issue #810, opened 2024-06-11
Define the Response Process for the Team to Handle Slack Notifications of Red Hat's LSP4IJ List of PRs


vaisakhkannan:
This issue is Part of Issue #802 ( Workitem 3)
We should define the response process for the team for handling these alerts. What actions do we need to take on our end? Is there a way to ensure that someone on the team takes these actions (i.e. re-alert the channel if no one completes the required action within a certain time frame?)

vaisakhkannan:
### Discussion was held with the Kochi Team regarding this Work Item 
Approach 1: Check the possibility of adding Yes or No buttons to the Slack notification. The Yes button should be pressed if someone in the team has visited the PR and reviewed the changes. This way, there will be no further reminders about the same PR in the Slack channel. If no one presses Yes, set a timeframe to send another notification reminding the channel about the same PR. Ensure to press the Yes button after viewing the PR and changes; otherwise, we will ignore the PR. It's uncertain if automating the reminder for each PR is achievable.

Approach 2: Set a daily reminder to visit the Red Hat repository's Open PR List at https://github.com/redhat-developer/lsp4ij/pulls to keep track of changes.

Approach 3: Workflow Builder: Slack's Workflow Builder enables you to create automated workflows without code. We can set up triggers and actions to automate processes like approvals, form submissions, and notifications.

vaisakhkannan:
### Found a way to trigger Slack’s Workflow Builder
There is an option in Slack called **Webhook**, which we need to use for creating the Slack Workflow Builder. After creation, we will receive a Webhook URL. We can use this Webhook URL within the workflow file of GitHub Actions (GHA). This will  trigger notifications in Slack for both Open and Closed PRs and triggering the created Slack Workflow Builder at the same time as the GHA workflow runs

vaisakhkannan:
### When I tried implementing  Step 2 and Step 3, I found some relevant points during investigations.

1. We can only set a reminder daily, or otherwise, we need to use multiple reminders set in Slack. That means we can’t set a reminder that starts at 9 AM ET and keeps going off every hour using Slack.
2. However, we can set another cron job file trigger with schedules based on starting at 9 AM ET and repeating every hour.

For example:
            on:
              schedule:
                - cron: '30 9-11 * * 1-5'
This schedule ensures that the workflow will run at 9:30 AM, 10:30 AM, and 11:30 AM on weekdays.

[...]

Issue #803, opened 2024-06-07
Determine How to Send Notifications to a Slack Channel for the List of PRs in the LSP4IJ Repo Based on Cron Jobs


vaisakhkannan:
Resource for Slack/GitHub integration: https://github.com/integrations/slack
This issue is specifically the second work item of issue #802

vaisakhkannan:
### The approach taken for the solution:

Step 1: Create a Slack App

Go to the Slack API and create a new app for your workspace.
Navigate to the "Incoming Webhooks" section and activate it.
Add the app to your desired channel to obtain the Webhook URL.
Step 2: Create a GitHub Actions Workflow

Create a Workflow File: In your GitHub repository, create a new file under the .github/workflows/ directory, e.g., .github/workflows/slack-notifications.yml.
Step 3: Use API calls to fetch information about new pull requests in the repository and send notifications to Slack.

Step 4: Add Slack Webhook Secret

In your GitHub repository, go to "Settings" -> "Secrets" and add a secret named SLACK_WEBHOOK_URL with the value of the Slack Webhook URL.
Step 5: Customize the GitHub API request in the workflow file based on our needs.

Step 6: Commit and Push Changes

Commit the new files (the workflow file) and push them to the GitHub repository.
Step 7: Verify Workflow Execution

After pushing the changes, check the Actions tab in our GitHub repository to verify that the workflow runs successfully.

vaisakhkannan:
### The solution used for this issue is a GitHub Actions workflow scheduled to run at 07:30 UTC (1:00 PM IST) on weekdays. It performs the following tasks
1. **Checkout Repository**: Ensures the repository is accessible.
2. **Fetch Recent Cache**: Installs the gh-actions-cache extension, retrieves the latest cache key, and sets it as an environment variable.
3. Restore Cache: Restores the cache using the fetched cache key.
4. Check Cache Hit: Verifies if the cache was restored successfully.
5. Ensure Cache Directory Exists: Creates a cache directory if it doesn't exist.
6. Check Cache Restoration: Confirms if the cache restoration was successful by checking for a specific file (cache/notified_prs.json)
[...]

PR #837, opened 2024-06-21
Fix for Sending Notification to Slack Channel on List of PRs in LSP4IJ

vaisakhkannan:
Fixes https://github.com/OpenLiberty/liberty-tools-intellij/issues/803
Added a new YAML file, `pr-slack-notification.yaml,` for sending Slack notifications based on cron job runs.

Fixes https://github.com/OpenLiberty/liberty-tools-intellij/issues/810

At the end of `pr-slack-notification.yaml`, I have added one more job called `slack-notification`, which is used to trigger Slack Workflow Builder as a fix for the issue https://github.com/OpenLiberty/liberty-tools-intellij/issues/810

### DIFF
--- a/.github/workflows/pr-slack-notification.yaml
+++ b/.github/workflows/pr-slack-notification.yaml
@@ -115,9 +115,9 @@ jobs:
 
           if [ "$notify" = true ]; then
             echo -e "$pr_list" > pr_list.txt
-            echo "::set-output name=notify::true"
+            echo "notify=true" >> $GITHUB_ENV
           else
-            echo "::set-output name=notify::false"
+            echo "notify=false" >> $GITHUB_ENV
           fi
 
       - name: Fetch closed pull requests since the last run
@@ -145,13 +145,13 @@ jobs:
           
           if [ -n "$closed_pr_list" ]; then
             echo -e "$closed_pr_list" > closed_pr_list.txt
-            echo "::set-output name=notify_closed::true"
+            echo "notify_closed=true" >> $GITHUB_ENV
           else
-            echo "::set-output name=notify_closed::false"
+            echo "notify_closed=false" >> $GITHUB_ENV
           fi      
 
       - name: Send Slack notification for PRs
-        if: success() && (steps.fetch_all_pull_requests_using_condition.outputs.notify == 'true' || steps.fetch_closed_prs.outputs.notify_closed == 'true')
+        if: success() && (${{ env.notify }} == 'true' || ${{ env.notify_closed }} == 'true')
         run: |
           # Initialize PR lists
           pr_list=""
--- a/.github/workflows/pr-slack-notification.yaml
+++ b/.github/workflows/pr-slack-notification.yaml
@@ -1,0 +1,304 @@
+name: Fetch All Open and Recently Closed Pull Requests using cron job
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '30 15 * * 1-5'  # At 15:30 on every day-of-week from Monday through Friday(UTC Time), 9pm in IST and 11:30 AM in EST.
+
+jobs:
+  fetch_all_pull_requests_and_notify_using_condition_and_cron:
+    runs-on: ubuntu-latest
+
+    env:
+      REPO: ${{ github.repository }}
+      BRANCH: main
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      API_URL: https://api.github.com/repos/redhat-developer/lsp4ij/pulls
+      SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Fetch recent cache
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          # Fetching list of cache keys........
+          cacheKeys=$(gh actions-cache list -R $REPO -B $BRANCH -L 100 | cut -f 1 )
+          
+          # Extract the most recent cache key
+          mostRecentCacheKey=$(echo "$cacheKeys" | head -n 1)
+          
+          if [ -n "$mostRecentCacheKey" ]; then
+            echo "CACHE_KEY=$mostRecentCacheKey" >> $GITHUB_ENV
+          else
+            echo "No recent cache key found, generating a new one...."
+            echo "CACHE_KEY=${{ runner.os }}-pr-cache-${{ github.run_number }}" >> $GITHUB_ENV
+          fi
+          
+          echo "Done"
+          # 'secrets.GITHUB_TOKEN' is automatically provided by GitHub Actions for each workflow run. You don't need to manually create or manage this token.
+
+      - name: Restore cache
+        id: cache-restore
+        uses: actions/cache@v2
+        with:
+          path: cache
+          key: ${{ env.CACHE_KEY }}
+
+      - name: Ensure cache directory exists and check if cache restored
+        run: |
+          mkdir -p cache
+          if [ -f cache/notified_prs.json ]; then
+            echo "Cache restored successfully."
+            cat cache/notified_prs.json
+          else
+            echo "Cache not restored or file does not exist.."
+          fi
+
+      - name: Fetch all opened pull request details using condition
+        id: fetch_all_pull_requests_using_condition
+        run: |
+          pr_infos=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          "${{ env.API_URL }}?state=open&sort=created&direction=desc&per_page=100")
+          
+          # Load previous PR data if exists
+          if [ -f cache/notified_prs.json ]; then
+            previous_prs=$(cat cache/notified_prs.json)
+          else
+            previous_prs="[]"
+          fi
+
+          pr_list=""
+          new_notified_prs="[]"
+          notify=false
+
+          for pr_info in $(echo "$pr_infos" | jq -r '.[] | @base64'); do
+            _jq() {
+              echo "$pr_info" | base64 --decode | jq -r "${1}"
+            }
+
+            pr_number=$(_jq '.number')
+            pr_title=$(_jq '.title')
+            pr_user=$(_jq '.user.login')
+            pr_url=$(_jq '.html_url')
+            pr_draft=$(_jq '.draft')
+            pr_created_at=$(_jq '.created_at')
+            pr_updated_at=$(_jq '.updated_at')
+
+            pr_data=$(jq -n --arg number "$pr_number" --arg updated_at "$pr_updated_at" '{number: $number, updated_at: $updated_at}')
+            new_notified_prs=$(echo "$new_notified_prs" | jq --argjson pr_data "$pr_data" '. += [$pr_data]')
+
+            echo "New Notified PR List------ $new_notified_prs"
+
+            # Check if the PR is new or updated
+            previous_pr=$(echo "$previous_prs" | jq --arg number "$pr_number" '.[] | select(.number == $number)')
+            echo "Checking PR is new or not --- $previous_pr"
+            if [ -z "$previous_pr" ] || [ "$(echo "$previous_pr" | jq -r '.updated_at')" != "$pr_updated_at" ]; then
+              draft_status=""
+              echo "------Draft value ---- $pr_draft"
+              if [ "$pr_draft" = "true" ]; then
+                draft_status="\n*Draft*: true"
+              fi
+
+              pr_list="${pr_list}\n*Pull Request* #${pr_number}: ${pr_title}\n*Created by*: ${pr_user}\n*URL*: ${pr_url}${draft_status}\n*Created At*: ${pr_created_at}\n*Last Updated At*: ${pr_updated_at}\n"
+              notify=true
+              echo "Boolean Value for notify ------- $notify"
+            fi
+          done
+
+          # Save current PR data for future comparison
+          echo "$new_notified_prs" > cache/notified_prs.json
+
+          if [ "$notify" = true ]; then
+            echo -e "$pr_list" > pr_list.txt
+            echo "::set-output name=notify::true"
+          else
+            echo "::set-output name=notify::false"
+          fi
+
+      - name: Fetch closed pull requests since the last run
+        id: fetch_closed_prs
+        run: |
+          if [ -f cache/last_run_timestamp.txt ]; then
+            last_run_timestamp=$(cat cache/last_run_timestamp.txt)
+          else
+            last_run_timestamp=$(date -u -d "-24 hours" +%Y-%m-%dT%H:%M:%SZ)
+          fi
+
+          echo "Last run timestamp: $last_run_timestamp"
+
+          closed_prs=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          "${{ env.API_URL }}?state=closed&sort=updated&direction=desc&per_page=100")
+
+          closed_pr_list=""
+
+          for pr_info in $(echo "$closed_prs" | jq -r '.[] | @base64'); do
+            _jq() {
+              echo "$pr_info" | base64 --decode | jq -r "${1}"
+            }
+
+            pr_number=$(_jq '.number')
+            pr_title=$(_jq '.title')
+            pr_user=$(_jq '.user.login')
+            pr_url=$(_jq '.html_url')
+            pr_closed_at=$(_jq '.closed_at')
+            pr_created_at=$(_jq '.created_at')
+
+            if [[ "$pr_closed_at" > "$last_run_timestamp" ]]; then
+              closed_pr_list="${closed_pr_list}\n*Closed Pull Request* #${pr_number}: ${pr_title}\n*Closed by*: ${pr_user}\n*URL*: ${pr_url}\n*Closed At*: ${pr_closed_at}\n"
+            fi
+          done
+          
+          echo "Latest Closed PR List ------$closed_pr_list"
+
+          if [ -n "$closed_pr_list" ]; then
+            echo -e "$closed_pr_list" > closed_pr_list.txt
+            echo "::set-output name=notify_closed::true"
+          else
+            echo "::set-output name=notify_closed::false"
+          fi      
+
+      - name: Send Slack notification for PRs
+        if: success() && (steps.fetch_all_pull_requests_using_condition.outputs.notify == 'true' || steps.fetch_closed_prs.outputs.notify_closed == 'true')
+        run: |
+          # Initialize PR lists
+          pr_list=""
+          closed_pr_list=""
+
+          # Check if the open PRs file exists and read its content
+          if [ -f pr_list.txt ]; then
+            pr_list=$(cat pr_list.txt)
+          fi
+
+          # Check if the closed PRs file exists and read its content
+          if [ -f closed_pr_list.txt ]; then
+            closed_pr_list=$(cat closed_pr_list.txt)
+          fi
+
+          # Initialize payload blocks
+          payload_blocks=()
+
+          # Add open PRs section if not empty
+          if [ -n "$pr_list" ]; then
+            payload_blocks+=("{
+              \"type\": \"header\",
+              \"text\": {
+                \"type\": \"plain_text\",
+                \"text\": \"List of Open/New/Updated Pull Requests using Cron Job\"
+              }
+            }")
+            payload_blocks+=("{
+              \"type\": \"section\",
+              \"text\": {
+                \"type\": \"mrkdwn\",
+                \"text\": \"$pr_list\"
+              }
+            }")
+          fi
+
+          # Add closed PRs section if not empty
+          if [ -n "$closed_pr_list" ]; then
+            payload_blocks+=("{
+              \"type\": \"header\",
+              \"text\": {
+                \"type\": \"plain_text\",
+                \"text\": \"List of Pull Requests Closed Since the Last Cron Job\"
+              }
+            }")
+            payload_blocks+=("{
+              \"type\": \"section\",
+              \"text\": {
+                \"type\": \"mrkdwn\",
+                \"text\": \"$closed_pr_list\"
+              }
+            }")
+          fi
+
+          # Construct the payload
+          payload=$(cat <<-EOF
+          {
+            "blocks": [
+              $(IFS=,; echo "${payload_blocks[*]}")
+            ]
+          }
+          EOF
+          )
+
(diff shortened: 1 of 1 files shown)


vaisakhkannan:
When I checked during testing in the cron job runs, I saw that fetching the Closed PR list was correct, but it was failing to send the payload. When I tested it locally, it was working. So I made a change to send it as a single payload that can be used for sending both open and closed PR lists in a single step in the workflow file. I hope it will work in today's cron jobs.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+            cat cache/notified_prs.json
+          else
+            echo "Cache not restored or file does not exist.."
+          fi
+
+      - name: Fetch all opened pull request details using condition
+        id: fetch_all_pull_requests_using_condition
+        run: |
+          pr_infos=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          "${{ env.API_URL }}?state=open&sort=created&direction=desc&per_page=100")
```
aparnamichael, line 64: sort=created, why are we sorting like this?
In lsp4ij PR's list, I am not seeing any value 'created' for sort parameter.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+        with:
+          path: cache
+          key: ${{ env.CACHE_KEY }}
+
+      - name: Ensure cache directory exists and check if cache restored
+        run: |
+          mkdir -p cache
+          if [ -f cache/notified_prs.json ]; then
+            echo "Cache restored successfully."
+            cat cache/notified_prs.json
```
aparnamichael, line 55: Do we need this cat command here?

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+            pr_user=$(_jq '.user.login')
+            pr_url=$(_jq '.html_url')
+            pr_draft=$(_jq '.draft')
+            pr_created_at=$(_jq '.created_at')
+            pr_updated_at=$(_jq '.updated_at')
+
+            pr_data=$(jq -n --arg number "$pr_number" --arg updated_at "$pr_updated_at" '{number: $number, updated_at: $updated_at}')
+            new_notified_prs=$(echo "$new_notified_prs" | jq --argjson pr_data "$pr_data" '. += [$pr_data]')
+
+            echo "New Notified PR List------ $new_notified_prs"
```
aparnamichael, line 93: Please remove unwanted echo statements.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+
+            # Check if the PR is new or updated
+            previous_pr=$(echo "$previous_prs" | jq --arg number "$pr_number" '.[] | select(.number == $number)')
+            echo "Checking PR is new or not --- $previous_pr"
+            if [ -z "$previous_pr" ] || [ "$(echo "$previous_pr" | jq -r '.updated_at')" != "$pr_updated_at" ]; then
+              draft_status=""
+              echo "------Draft value ---- $pr_draft"
```
aparnamichael, line 100: Please remove unwanted echo statements.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+          if [ -f cache/last_run_timestamp.txt ]; then
+            last_run_timestamp=$(cat cache/last_run_timestamp.txt)
+          else
+            last_run_timestamp=$(date -u -d "-24 hours" +%Y-%m-%dT%H:%M:%SZ)
+          fi
+
+          echo "Last run timestamp: $last_run_timestamp"
+
+          closed_prs=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          "${{ env.API_URL }}?state=closed&sort=updated&direction=desc&per_page=100")
```
aparnamichael, line 133: why are we using this parameter "sort=updated"?

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+
+          closed_pr_list=""
+
+          for pr_info in $(echo "$closed_prs" | jq -r '.[] | @base64'); do
```
aparnamichael, line 137: These lines of code are present in step "Fetch all opened pull request details using condition" and here. Is there is any way to avoid this code duplication?

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,304 @@
...
+          payload=$(cat <<-EOF
+          {
+            "blocks": [
+              $(IFS=,; echo "${payload_blocks[*]}")
+            ]
+          }
+          EOF
+          )
+
+          # Debugging: Print payload to verify its format
```
aparnamichael, line 229: Remove print statements used for debugging.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: This is just for printing to ensure that the cache was fetched correctly, I hope this will help in the future to check the fetched data correctly. If required, we can delete it.

Screenshot 2024-06-25 at 2 17 40 PM

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: If we comment out this echo statement, it will be difficult to understand the workflow logs in the future. We won't be able to debug this process. The only way to analyze the workflow is by using echo statements. If we remove them now, we might need to modify the code later when checking the process (including necessary changes like secrets).We may want to rerun the workflow which will goes repeated unwanted noice in slack channel. This could disrupt the current working code. This makes it challenging for other developers, especially those who have not worked on this previously. Clear workflow logs help easily identify issues.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: It's challenging to avoid code duplication by creating a reusable step or script that handles the common functionality between 'Fetch closed pull requests since the last run' and 'Fetch all opened pull request details using conditions' because both steps involve different logics. Additionally, we may require additional conditional checks, which complicates efforts to prevent code duplication.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: Yes I removed it.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: Yes I modified the code.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: Yes I modified it

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: Yes I modified it.

.github/workflows/pr-slack-notification.yaml
aparnamichael: @vaisakhkannan  Yes, logs help us understand the results at each step, but adding more logs can affect the performance especially inside the iterative statements like for/while loops and it will make. Also it will reduce the readability . For example, the screenshot you have attached shows logs for each PR detail with "Checking PR is new or not," followed by printing the list again with "New Notified PR List."

Instead, you could simply log the list of PRs retrieved by the API and the list of new or updated PRs, right?

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: The GitHub API expects certain values for the state parameter, such as `open, closed, or all`, but it does not support combining states like `(closed || open)` directly.

If we want to fetch both open and closed pull requests, we can use `state=all` to include all pull requests regardless of their state. However, this approach has drawbacks. It includes all PRs (both open and closed) in the comparison with the `cached PR list`, which increases the time and computational requirements. Additionally, using `state=all` prevents us from specifying a limit like `per_page=100`, as it will fetch all PRs in the repository (currently around 384 PRs in lisp4ij).

Screenshot 2024-06-26 at 10 23 57 AM

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,297 @@
...
+          
+          if [ -n "$mostRecentCacheKey" ]; then
+            echo "CACHE_KEY=$mostRecentCacheKey" >> $GITHUB_ENV
+          else
+            echo "No recent cache key found, generating a new one...."
+            echo "CACHE_KEY=${{ runner.os }}-pr-cache-${{ github.run_number }}" >> $GITHUB_ENV
+          fi
+          
+          echo "Done"
+          # 'secrets.GITHUB_TOKEN' is automatically provided by GitHub Actions for each workflow run. You don't need to manually create or manage this token.
```
mrglavas, line 41: Is this comment in the right place? It doesn't seem related to the line of code immediately above it.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,297 @@
...
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
+
+      - name: Fetch recent cache
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          # Fetching list of cache keys........
+          cacheKeys=$(gh actions-cache list -R $REPO -B $BRANCH -L 100 | cut -f 1 )
```
mrglavas, line 28: I guess this is limiting the list of cache keys returned to 100. Are we likely to hit this limit? What would happen if we do?

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,297 @@
...
+            cat cache/notified_prs.json
+          else
+            echo "Cache not restored or file does not exist.."
+          fi
+
+      - name: Fetch all opened pull request details using condition
+        id: fetch_all_pull_requests_using_condition
+        run: |
+          pr_infos=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          "${{ env.API_URL }}?state=open&direction=desc&per_page=100")
```
mrglavas, line 64: If I'm reading this correctly it looks like this is limited to fetching 100 open PRs (per page). Hopefully there are never that many open PRs but I'm curious if this code would work if there's more than 100.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,297 @@
...
+          if [ -f cache/last_run_timestamp.txt ]; then
+            last_run_timestamp=$(cat cache/last_run_timestamp.txt)
+          else
+            last_run_timestamp=$(date -u -d "-24 hours" +%Y-%m-%dT%H:%M:%SZ)
+          fi
+
+          echo "Last run timestamp: $last_run_timestamp"
+
+          closed_prs=$(curl -s -H "Authorization: token ${{ secrets.GITHUB_TOKEN }}" \
+          "${{ env.API_URL }}?state=closed&direction=desc&per_page=100")
```
mrglavas, line 131: Similar question about the limit on the number closed PRs. Would this work if there's more than 100?

.github/workflows/pr-slack-notification.yaml
TrevCraw: FYI for the scenario of this workflow, performance is not very important. Readability is highly important, though.

.github/workflows/pr-slack-notification.yaml
TrevCraw: @vaisakhkannan I noticed you removed `sort=updated` without explaining what it was for. Was there a reason you had that parameter initially?

vaisakhkannan: @TrevCraw Initially, I thought it would be better to sort the closed PRs based on `sort=updated`. However, then I had realized that if we ran daily cron jobs, there was no need to sort, and the list may be within the 100 limit. If the number of closed PRs exceeds this limit, then we will definitely need `sort=updated`. So now I think based on Michael's comment aswell, I have added back to using `sort=updated`. In the future, we expect to have more than 100 closed PRs."

Screenshot 2024-06-27 at 4 28 11 PM

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: By default is `sort=created`. So I removed this. If require we can add it back.
 https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @aparnamichael, @TrevCraw , I modified the code to list currently Opened PRs and List of PRs that need to be sent to Slack after that current run

Screenshot 2024-06-27 at 4 57 56 PM

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @aparnamichael , I simplified the code for the step `'Fetch closed pull requests since the last run'`, and I attempted the same approach to simplify the code for the step `'Fetch all opened pull request details using conditions'` as well. I avoided using a for loop and instead utilized jq with select queries to filter out values. However, when checking the conditions we implemented for listing PRs (specifically, `omitting PRs already listed in Slack or those that have been updated or are new`), it did not succeed. If necessary, I will investigate further.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @mrglavas , Allowed values are within this limit `1-100` , Only the latest 100 cache keys will be listed. Any older cache keys beyond this limit will not be included in` $cacheKeys`. I am hoping that last cron job key will not be beyond the limit of 100.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @mrglavas , During my investigation , I found that "The number of results per page (max 100)". We  [pagination](https://docs.github.com/en/rest/using-the-rest-api/using-pagination-in-the-rest-api?apiVersion=2022-11-28) , if we need to increase the limit.

With this code, PR list will be limited to 100 and we can sort PRs based on creation date by "sort=created" .

.github/workflows/pr-slack-notification.yaml
mrglavas: @vaisakhkannan It sounds unlikely that we would hit the limit but it would be good to document that the workflow has this limitation.

vaisakhkannan: @mrglavas, I added the comment in the code

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,287 @@
...
+            gh actions-cache delete $mostRecentCacheKey -R $REPO -B $BRANCH --confirm
+          else
+            echo "No cache keys found."
+          fi
+
+          echo "Done"
+
+    # 'secrets.GITHUB_TOKEN' is automatically provided by GitHub Actions for each workflow run. We don't need to manually create or manage this token.
+    # The below way, the workflow is granting write access to the actions scope of the GITHUB_TOKEN.
+    # If the below specific way is not successful to delete the cache key, then the repository owner will need to create a Personal Access Token (PAT) with write permissions, add it as a secret, and specify the name in GH_TOKEN.
```
TrevCraw, line 262: I am confused by the location of these comments

.github/workflows/pr-slack-notification.yaml
TrevCraw: @vaisakhkannan Thank you for simplifying the code where able. If it is becoming too complicated or too complex to simplify the code elsewhere, there is no need to do so.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @TrevCraw , I added the comment for the below code.
```
permissions:
      actions: write
```

I mentioned this in the box note that this code is responsible for deleting the `cache key`. If we do not specify it this way, the deletion of the cache key cannot happen, and we will need a PAT, which needs to be added as a secret for deletion.

Screenshot 2024-06-29 at 11 03 15 AM

For reference : https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defining-access-for-the-github_token-scopes-1

If require, we can modify the comment.

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,287 @@
...
+          else
+            echo "Cache not restored or file does not exist.."
+          fi
+
+      - name: Fetch all opened pull request details using condition
+        id: fetch_all_pull_requests_using_condition
+        run: |
+          # The number of results per page is limited to a maximum of 100. We need "pagination", if we require more than this limit, but for open PRs, it will be within the limit 100.
+          pr_infos=$(curl -s -H "Authorization: token ${{ env.GH_TOKEN }}" \
+          "${{ env.API_URL }}?state=open&direction=desc&per_page=100")
```
anusreelakshmi934, line 65: Is it necessary to specify state=open&direction=desc since the default values are state=open and direction=desc?

vaisakhkannan: I believe that, its good to keep `state=open@direction=desc` , because the default may change in future, so it will break the code when we try to fetch Open PRs

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,287 @@
...
+          
+          echo "List of PRs that need to be sent to Slack now:"
+          echo "$pr_list"
+
+          # Save current PR data for future comparison
+          echo "$new_notified_prs" > cache/notified_prs.json
+
+          if [ "$notify" = true ]; then
+            echo -e "$pr_list" > pr_list.txt
+            echo "::set-output name=notify::true"
```
anusreelakshmi934, line 118: There is a warning in the workflow file - `The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files.`. So it is better to change set-output to the GITHUB_OUTPUT environment files. Ref - https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

aparnamichael:
@vaisakhkannan Can you please share the link of one workflow run?

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,287 @@
...
+            pr_number=$(_jq '.number')
+            pr_title=$(_jq '.title')
+            pr_user=$(_jq '.user.login')
+            pr_url=$(_jq '.html_url')
+            pr_draft=$(_jq '.draft')
+            pr_created_at=$(_jq '.created_at')
+            pr_updated_at=$(_jq '.updated_at')
+
+            pr_data=$(jq -n --arg number "$pr_number" --arg updated_at "$pr_updated_at" '{number: $number, updated_at: $updated_at}')
+            new_notified_prs=$(echo "$new_notified_prs" | jq --argjson pr_data "$pr_data" '. += [$pr_data]')
```
aparnamichael, line 95: What do we store in 'pr_data' variable? 
Why can't we directly parse and assign value to 'new_notified_prs' variable?

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,287 @@
...
+
+            # Check if the PR is new or updated
+            previous_pr=$(echo "$previous_prs" | jq --arg number "$pr_number" '.[] | select(.number == $number)')
+            if [ -z "$previous_pr" ] || [ "$(echo "$previous_pr" | jq -r '.updated_at')" != "$pr_updated_at" ]; then
```
aparnamichael, line 99: Why are we checking these conditions for getting draft status? In can see from the api itself we are getting the draft status.

.github/workflows/pr-slack-notification.yaml
TrevCraw: @vaisakhkannan Thank you for the explanation. We should make it clear that the write permission enables the cache key to be deleted. I would recommend the below modifications:
```suggestion
    # Below, the workflow is granting "write" access to the actions scope of the GITHUB_TOKEN. This enables the workflow to be able to delete cache keys.
    # If the "write" permission does not enable the ability to delete the cache keys, then the repository owner will need to create a Personal Access Token (PAT) with write permissions, add it as a secret, and then specify the name in GH_TOKEN.
```

vaisakhkannan:
@aparnamichael,  The link of workflow run --> https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/9780833228/job/27003430709

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @aparnamichael , Yes, when I tried to directly parse and assign the value to 'new_notified_prs', it made it difficult to read.

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @anusreelakshmi934 , I modified the code to use `$GITHUB_ENV` , because this is using for pass values from one step to another step within the same job. Now I am not getting the warning.

Example of recent run: https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/9793149679/job/27040487378

aparnamichael (dismissed):
Looks good!

anusreelakshmi934 (approved):
Looks Good. Thanks

.github/workflows/pr-slack-notification.yaml
```diff
@@ -0,0 +1,289 @@
...
+            gh actions-cache delete $mostRecentCacheKey -R $REPO -B $BRANCH --confirm
+          else
+            echo "No cache keys found."
+          fi
+
+          echo "Done"
+
+    # 'secrets.GITHUB_TOKEN' is automatically provided by GitHub Actions for each workflow run. We don't need to manually create or manage this token.
+    # The below way, the workflow is granting write access to the actions scope of the GITHUB_TOKEN.
+    # Below, the workflow is granting "write" access to the actions scope of the GITHUB_TOKEN. This enables the workflow to be able to delete cache keys.
```
TrevCraw, line 263: Line 263 was intended to replace 262

TrevCraw (changes requested):

.github/workflows/pr-slack-notification.yaml
vaisakhkannan: @TrevCraw , I removed the line 262.

aparnamichael (approved):

mrglavas (approved):

TrevCraw (approved):

PR #853, opened 2024-07-04
Fix for Liberty Start issue in directory with space in name 

### LINKED ISSUE #690: Liberty Start in directory with space in name
When you use a project with a space in the name (e.g. "GrdEE91 app") then Liberty Start does not work:
image

The cd has quotes but the gradlew needs quotes too.

vaisakhkannan:
Fixes #690 
Fixed `Liberty Start` issue in directories with spaces in their names. During my investigation, I found that the problem occurs only with Gradle projects, and there are no issues with Maven projects. I wrote tests for Gradle projects that cover cases with `spaces in app names and directory names`. Similarly, I wrote tests for Maven projects that cover cases with spaces in directory names. We can't use spaces in app names with `Maven projects`, as this will result in an error in the pom.xml file stating that `the value of the artifactId does not match the id pattern`.

Screenshot 2024-07-01 at 4 21 05 PM

The problem will occur with both `Liberty: Start and Liberty: Start in Container`. I have written tests only for `Liberty: Start`. If required, I can add tests for the other scenario.

I have added a `new sample project` because we can't use the existing projects. We need a project with a space in its name (only for Gradle) and a space in the directory name. So, I created another directory with a space in its name.

I have checked the tests to see whether they will fail (based on whether Liberty starts or not). If the bug in the issue still exists, the gradle test will fail(`No such file or directory error`, resulting in `Liberty:Start `action failure, for maven there is no issue with the directory that getting ,and the functionality in `gradle wrapper` and `maven wrapper` are slightly different. In maven wrapper we get a command like `"./mvnw"` but in gradle wrapper , it will get path of the project aswell , that's the reason behind the failure in `gradle` only), otherwise, it will succeed. 

Screenshot 2024-07-02 at 1 49 59 PM

Test failure in Maven projects depends on how the `Liberty:Start` action performs. If `Liberty:Start` fails, it will result in a failure error.

The failure results in both the cases are same as shown in the screenshot above


src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
```diff
@@ -230,7 +230,8 @@ private static String getLocalGradleWrapperPath(String wrapperDir) throws Libert
         } catch (IOException e) {
             throw new LibertyException("Could not get canonical path for gradle wrapper file");
         }
-        return path;
+        String additionalCMD = SystemInfo.isWindows ? "cmd /K " : ""; // In Windows, "cmd /K" is used to execute a command
+        return additionalCMD + LibertyProjectUtil.includeEscapeToString(path);
```
turkeylurkey, line 234: I see where you got this code but I find these two lines too complex. Also we have to do an unnecessary concatenation operation on line 234 for mac and linux. How about a more conventional:
String cmd = LibertyProjectUtil.includeEscapeToString(path);
if (SystemInfo.isWindows) {
  cmd = "cmd /K " + cmd;
}
return cmd;

vaisakhkannan: Thanks @turkeylurkey for the suggestion , I modified the code.

src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
```diff
@@ -230,7 +230,10 @@ private static String getLocalGradleWrapperPath(String wrapperDir) throws Libert
         } catch (IOException e) {
             throw new LibertyException("Could not get canonical path for gradle wrapper file");
         }
-        return path;
+        String cmd = LibertyProjectUtil.includeEscapeToString(path);
+        if (SystemInfo.isWindows)
```
aparnamichael, line 234: Can we use ternary operator here similar to line no. 254, I felt the below code is more readable.

 String additionalCMD = SystemInfo.isWindows ? "cmd /K " : ""; // without it, a new terminal window is opened
 return additionalCMD + LibertyProjectUtil.includeEscapeToString(gradleExecutable.getAbsolutePath());

aparnamichael:
@vaisakhkannan Please modify the year in header of all the modified files.

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
```diff
@@ -986,6 +991,35 @@ public void testCustomStartParametersClearedOnConfigRemoval() {
...
+            TestUtils.deleteDirectory(dir);
+        }
+    }
+
+    /**
+     * Returns the gradle project directory path with space in directory name
+     *
+     * @return The projects directory path.
+     */
+    private static String getGradleProjectSIDDirPath() {
```
aparnamichael, line 1011: There is already methods to get the directory path 'getProjectsDirPath'. Then why do we need this and getMavenProjectSIDDirPath method?

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
```diff
@@ -68,6 +69,10 @@ public static void cleanup() {
         UIBotTestUtils.closeProjectView(remoteRobot);
         UIBotTestUtils.closeProjectFrame(remoteRobot);
         UIBotTestUtils.validateProjectFrameClosed(remoteRobot);
+
+        // Delete the Gradle and Maven project directories with spaces in their names that were created during the environment setup for tests.
+        deleteDirectoryIfExists(getGradleProjectSIDDirPath());
```
aparnamichael, line 74: Can we move these lines of codes to Gradle and Maven classes. 
Once the Gradle tests are completed, you can delete the Gradle project. Similarly, after the Maven project is finished, you can delete the Maven project as well.

src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
vaisakhkannan: @aparnamichael  ,Paul suggested this change. Initially, I made the same change you are suggesting now. Then Paul provided this suggestion as a fix. He already have raised a PR for changing the code in line number 254

https://github.com/OpenLiberty/liberty-tools-intellij/pull/871/files

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
vaisakhkannan: `getProjectsDirPath()` is an abstract method, and we can't use it in the `cleanup()` method, which is annotated with `@ AfterAll` . 
We only need to call the method `deleteDirectoryIfExists()` if the path string is:

```
Paths.get("src", "test", "resources", "projects", "gradle sample").toAbsolutePath().toString();
Paths.get("src", "test", "resources", "projects", "maven sample").toAbsolutePath().toString();
```
We don't need to call the method `deleteDirectoryIfExists()` if the path string is:


```
Paths.get("src", "test", "resources", "projects", "gradle").toAbsolutePath().toString();
Paths.get("src", "test", "resources", "projects", "maven").toAbsolutePath().toString();
```
Since we are using an inheritance approach, do we need to change the following method to non-abstract?


`public abstract String getProjectsDirPath();`

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
vaisakhkannan: @aparnamichael , II initially tried the same approach, but the directory is still present after the corresponding Gradle and Maven tests are completed. When I debugged, I found that the cleanup() method in this class is executing last. It uses four methods (lines 68-71) that utilize the project directory path, which is causing the directory to remain after the tests are completed.

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
aparnamichael: Why can't we write the after all method inside GradleSingleModMPSIDProjectTest and MavenSingleModMPSIDProjectTest?

    @AfterAll
    public static void cleanup() {
        deleteDirectoryIfExists(PROJECTS_PATH_NEW);
    }

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
aparnamichael: My understanding is lines 68-71 is not using directory path, they are trying to find the UI element to close the project.  I moved the deleteDirectoryIfExists() to the GradleSingleModMPSIDProjectTest and MavenSingleModMPSIDProjectTest @AfterAll method. And the Test was working fine. Please try the same.

vaisakhkannan: I tried the same way you suggested, but the UI test is not completing all the actions. Specifically, the code below is not executing.

```
@AfterAll
public static void cleanup() {
    UIBotTestUtils.closeLibertyToolWindow(remoteRobot);
    UIBotTestUtils.closeProjectView(remoteRobot);
    UIBotTestUtils.closeProjectFrame(remoteRobot);
    UIBotTestUtils.validateProjectFrameClosed(remoteRobot);
}
```

The tests are stopping after executing the code we wrote in the GradleSingleModMPSIDProjectTest and MavenSingleModMPSIDProjectTest classes.

```
@AfterAll
public static void cleanup() {
    deleteDirectoryIfExists(PROJECTS_PATH_NEW);
}
```

aparnamichael: Can we change the code like this.

/**
 * Cleanup.
 */
@AfterAll
public static void cleanup() {
    closeProjectView();
}

protected static void closeProjectView() {
    // Parent cleanup logic
    UIBotTestUtils.closeLibertyToolWindow(remoteRobot);
    UIBotTestUtils.closeProjectView(remoteRobot);
    UIBotTestUtils.closeProjectFrame(remoteRobot);
    UIBotTestUtils.validateProjectFrameClosed(remoteRobot);
}

child class

@AfterAll
public static void cleanup() {
    closeProjectView();
    deleteDirectoryIfExists(PROJECTS_PATH_NEW);
}

vaisakhkannan: @aparnamichael , I have updated in the PR changes as per your suggestions.

aparnamichael (dismissed):
Looks good!

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
```diff
@@ -0,0 +1,252 @@
...
+        Path originalPathCopy = projectDirPath.resolve("settings-copy.gradle");
+
+        // Rename settings.gradle to settings-duplicate.gradle
+        Files.move(originalPath, originalPath.resolveSibling("settings-duplicate.gradle"));
+        // Rename settings-copy.gradle to settings.gradle
+        Files.move(originalPathCopy, originalPathCopy.resolveSibling("settings.gradle"));
+
+        Path projectDirNewPath = pathNew.resolve(SM_MP_PROJECT_NAME_NEW);
+
+        try {
```
turkeylurkey, line 114: I don't understand the exception handling strategy here. Do you really just want to print a message when the directory change is made but not on other errors? It seems like the three lines above this should also be in this try statement and the error message adjusted. You probably should also use an assertion to terminate the test if there is an exception.

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
```diff
@@ -0,0 +1,252 @@
...
+     */
+    private final String DEV_MODE_START_PARAMS_DEBUG = "--libertyDebugPort=9876";
+
+    /**
+     * Prepares the environment for test execution.
+     */
+    @BeforeAll
+    public static void setup() throws IOException {
+        // Copy the directory from PROJECTS_PATH to PROJECTS_PATH_NEW
+        TestUtils.copyDirectory(PROJECTS_PATH, PROJECTS_PATH_NEW);
```
turkeylurkey, line 98: If you handle the IOException from this line you can remove the `throws` definition above (in combination with the changes below.

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
```diff
@@ -0,0 +1,252 @@
...
+        Files.move(originalPath, originalPath.resolveSibling("settings-duplicate.gradle"));
+        // Rename settings-copy.gradle to settings.gradle
+        Files.move(originalPathCopy, originalPathCopy.resolveSibling("settings.gradle"));
+
+        Path projectDirNewPath = pathNew.resolve(SM_MP_PROJECT_NAME_NEW);
+
+        try {
+            // Rename the project directory to a new name, replacing it if it already exists
+            Files.move(projectDirPath, projectDirNewPath, StandardCopyOption.REPLACE_EXISTING);
+            System.out.println("Directory renamed successfully");
```
turkeylurkey, line 117: I don't think we need a message when the operation is successful.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -540,5 +540,25 @@ public static void checkDebugPort(String absoluteWLPPath, int debugPort) throws
...
+     * @param destinationDirectoryLocation The path to the destination directory to copy to.
+     * @throws IOException If an I/O error occurs while copying files.
+     */
+    public static void copyDirectory(String sourceDirectoryLocation, String destinationDirectoryLocation)
+            throws IOException {
+        Files.walk(Paths.get(sourceDirectoryLocation))
+                .forEach(source -> {
+                    Path destination = Paths.get(destinationDirectoryLocation, source.toString()
+                            .substring(sourceDirectoryLocation.length()));
+                    try {
```
turkeylurkey, line 556: Why do we need to catch this exception but not the one from `Files.walk()`? The test can't proceed if any files are missing and if it does a strange failure will result. You should remove this `try/catch`.

vaisakhkannan:
@aparnamichael , @turkeylurkey This is a screen recording showing how the exception occurs when trying to access an invalid file.

https://github.com/user-attachments/assets/bbb35c4d-592e-4d7c-82c7-9a554540032a

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
vaisakhkannan: I made the changes according to your suggestions and tested them. Everything was successful.

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
vaisakhkannan: Yes I removed the line.

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
vaisakhkannan: If an exception occurs within the `setup()` method, it will result in an initialization error because the method will not execute `prepareEnv(PROJECTS_PATH_NEW, SM_MP_PROJECT_NAME_NEW`). Consequently, all tests in the parent class `SingleModMPProjectTestCommon` will not work, leading to an` initialization error`.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
vaisakhkannan: @turkeylurkey , I found this code from [here ](https://www.baeldung.com/java-copy-directory#using-the-javanio-api) 
I tried removing the `try-catch` block from here and adding it from `Files.walk()`, but I'm getting an error on this line. The quick-fix suggests adding a `try-catch` block for the following line: `Files.copy(source, destination);`

Screenshot 2024-08-16 at 5 35 48 PM


Since we throw an `IOException` for the `copyDirectory(String sourceDirectoryLocation, String destinationDirectoryLocation)` method, I believe that the exception will definitely be caught if any I/O operation fails and I believe the actual operation is performing at this line `Files.copy(source, destination);`

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPSIDProjectTest.java
```diff
@@ -0,0 +1,258 @@
...
+
+    /**
+     * Cleanup includes deleting the created project path.
+     */
+    @AfterAll
+    public static void cleanup() {
+        try {
+            closeProjectView();
+        } finally {
+            deleteDirectoryIfExists(PROJECTS_PATH_NEW);
```
vaisakhkannan, line 136: I have added this method in the finally block because it must execute even if an exception occurs. This ensures that the newly created directory is removed, leading to complete cleanup.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: I believe an IOException on line 557 will be caught and this method will complete normally with just a stack trace as evidence there was an error. I think it would be better if the test failed. Could you throw a RuntimeException or some other type to indicate this error when you catch IOException?

vaisakhkannan: @turkeylurkey, I made the change , please have a look

turkeylurkey: This looks good but have you tested it? Could you mock a failed file copy and show us what the results and the logs look like?

vaisakhkannan: @turkeylurkey , I am able to reproduce the exception by adding the same line `Files.copy(source, destination)`  again as shown in the image below

Screenshot 2024-08-29 at 11 37 27 AM

aparnamichael (approved):

turkeylurkey (approved):
Ok, thanks for the explanation.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: 👍

PR #862, opened 2024-07-09
Use different Slack workflow builder message when there are no new/Updated/Closed PRs

### LINKED ISSUE #848: Investigate How to Send a Slack Message to Channel if There Are No New, Updated, or Closed PRs in the Current Run.
Currently, there is a similar message in the Slack workflow that triggers every cron job, even if there are no new, updated, or closed PRs to list.

Message displaying now as shown in the image

Screenshot 2024-06-28 at 12 12 14 PM

vaisakhkannan:
Fixes #848 

I created a new `Slack Workflow Builder(No LSP4IJ PRs)` for sending messages to a Slack channel when there are **no new, closed, or updated PRs** to list(Option 2 that mentioned in issue #848) . I added the new webhook URL from Slack Workflow Builder as a secret in my forked Intellij repository.

### DIFF
--- a/.github/workflows/pr-slack-notification.yaml
+++ b/.github/workflows/pr-slack-notification.yaml
@@ -15,6 +15,8 @@ jobs:
       GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       API_URL: https://api.github.com/repos/redhat-developer/lsp4ij/pulls
       SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      WORKFLOW_BUILDER_WEBHOOK: ${{ secrets.WORKFLOW_BUILDER_WEBHOOK }}
+      NO_PR_WORKFLOW_BUILDER_WEBHOOK: ${{ secrets.NO_PR_WORKFLOW_BUILDER_WEBHOOK }}
 
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


vaisakhkannan:
Result:
Screenshot 2024-07-09 at 6 53 08 PM

Reference cron job link: https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/9851228841/job/27197735484

.github/workflows/pr-slack-notification.yaml
```diff
@@ -258,20 +260,8 @@ jobs:
 
           echo "Done"
 
-    # 'secrets.GITHUB_TOKEN' is automatically provided by GitHub Actions for each workflow run. We don't need to manually create or manage this token.
-    # Below, the workflow is granting "write" access to the actions scope of the GITHUB_TOKEN. This enables the workflow to be able to delete cache keys.
-    # If the "write" permission does not enable the ability to delete the cache keys, then the repository owner will need to create a Personal Access Token (PAT) with write permissions, add it as a secret, and then specify the name in GH_TOKEN.
-    permissions:
-      actions: write
-
-  slack-notification:
```
aparnamichael, line 267: Previously, Slack notification was a separate job, why it became part of this job now ?

vaisakhkannan: @aparnamichael , I am passing values between different steps of the same job `"fetch_all_pull_requests_and_notify_using_condition_and_cron"` using `GITHUB_ENV` (GitHub Environment variable). To pass values between different jobs, I need to use artifacts or `outputs`. However, in this case, I require both outputs and GITHUB_ENV, which would lead to duplication. GITHUB_ENV is needed for passing values between steps within one job, while outputs are required for passing values between jobs. Therefore, I thought it would be better to avoid adding another job and instead keep it as an additional step. The execution of that step will be based on the values from GITHUB_ENV (env.notify, env.notify_closed). If there is no condition to check for the execution, then we can add it as another job, but here that is not applicable.

aparnamichael (approved):
Looks good!!

mrglavas (approved):
Looks good. Thanks.

TrevCraw (approved):

PR #871, opened 2024-07-11
Refactor gradle command generator to return at the end of the method

### LINKED ISSUE #869: Refactor the code for gradle command
Make the command processing method returns at the bottom of the method. This means checking for errors first.

turkeylurkey:
The old code was used as inspiration for new code so this should be cleaned up to be more readable.

Fixes #869

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2022 IBM Corporation.
+ * Copyright (c) 2020, 2024 IBM Corporation.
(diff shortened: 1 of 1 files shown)


TrevCraw:
Hi @turkeylurkey, I am not clear on the motivation here. Is it just readability?

TrevCraw:
Hi @turkeylurkey, I am not clear on the motivation here. Is it just readability?

turkeylurkey:
It is not just readability. Someone copied the code from here to another method so I wanted to improve readability so when it is copied the new methods will be better too.

turkeylurkey:
Also other methods in the class do return at the end and use if statements as structured in this pull request so another good reason to merge this.

TrevCraw (approved):

mrglavas (approved):
Looks good. Thanks.

PR #900, opened 2024-08-05
Changed Cron job execution time

anusreelakshmi934:
Cron job to runs at 15:45 UTC every Monday through Friday, which is 9:15 PM IST and 11:45 AM EST.


vaisakhkannan (approved):
looks good now. Thanks.

aparnamichael (approved):

Issue #905, opened 2024-08-07
Use logger to report exceptions instead of directly calling printStackTrace().


mrglavas:
There are several places we're calling `Throwable.printStackTrace()` in `io/openliberty/tools/intellij/actions/ViewTestReport.java`. We should clean this up.

PR #907, opened 2024-08-08
Write exceptions to the logger instead of directly printing their stack trace to the console.

mrglavas:
Resolves https://github.com/OpenLiberty/liberty-tools-intellij/issues/905.


aparnamichael:
@mrglavas I noticed that `printStackTrace()` is being used in the `TestUtils.java` and `UIBotTestUtils.java` classes. Should we need to update those as well?

mrglavas:
> @mrglavas I noticed that `printStackTrace()` is being used in the `TestUtils.java` and `UIBotTestUtils.java` classes. Should we need to update those as well?

@aparnamichael The concern is with production code, classes which are included in the plug-in that we publish to the JetBrains Marketplace. `TestUtils.java` and `UIBotTestUtils.java` are test classes. They are not shipped so the `printStackTrace()` calls are okay there.

aparnamichael (approved):
Looks good!!

turkeylurkey (approved):

PR #908, opened 2024-08-12
Fix warnings in Workflows on macOS 14.6 for curl 

### LINKED ISSUE #902: Warnings in GitHub Workflows on macOS 14.6 for curl Version 8.9.1
I found the following warning in Mac builds (specifically in the step to install required integration test software).

`curl 8.9.1 is already installed and up-to-date. To reinstall 8.9.1, run: brew reinstall curl`

Screenshot 2024-08-06 at 8 06 56 PM

When I checked this issue, I found that this warning occurs only on GHA Mac machines with OS version 14.6 and not on 14.5.

Here is an example of a workflow run: https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/10266055640
(  Warning came only in the `main` branch, PR number `19`, and PR number `11`. Those Mac builds were run on macOS version `14.6` and others are were run on Mac OS version `14.5`)

vaisakhkannan:
Fixes #902 

Since we are using `macOS-latest`, which refers to `macOS 14 Arm64`,we can check [here](https://github.com/actions/runner-images?tab=readme-ov-file#about)

Upon reviewing the [readme](https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md), I found that they are using `macOS version 14.6.1` now, and this update is reflected in the `documentation`. Therefore, we can confirm that they won't revert to `macOS version` 14.5. 

I have suppressed the warning for the command `brew install curl unzip || true` by using the` --quiet` option. 
The `--quiet` option in Homebrew's `brew install` command is used to minimize the output by suppressing some of the usual messages that appear during installation.

I have updated the branch to` lsp4ij-market-0.3.0-integration` in [this line](https://github.com/OpenLiberty/liberty-tools-intellij/blob/8287c40a55a83c8dc0a6d2081fbacd7a8a59a971/.github/workflows/build.yaml#L45) to enable tests on the feature branch.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,7 +42,7 @@ on:
   push:
     branches: '**'
   pull_request:
-    branches: [ main, lsp4ij-market-0.0.2-integration ]
+    branches: [ main, lsp4ij-market-0.3.0-integration ]
 
 jobs:
   build:
--- a/src/test/resources/ci/scripts/setup.sh
+++ b/src/test/resources/ci/scripts/setup.sh
@@ -93,7 +93,7 @@ installBaseSoftware() {
         installDockerOnLinux
     elif [[ $OS == "Darwin" ]]; then
         brew update
-        brew install curl unzip || true
+        # brew install curl unzip || true
         # installDockerOnMAC
     else
         # Note: Docker is already installed on the windows VMs provisioned by GHA.
--- a/src/test/resources/ci/scripts/setup.sh
+++ b/src/test/resources/ci/scripts/setup.sh
@@ -93,7 +93,7 @@ installBaseSoftware() {
         installDockerOnLinux
     elif [[ $OS == "Darwin" ]]; then
         brew update
-        # brew install curl unzip || true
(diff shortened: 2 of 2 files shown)


vaisakhkannan:
Based on the discussion with Trevor , we may use other macOS platforms to run tests on GHA in the future. Some earlier versions of macOS do not have curl installed by default or may have an outdated version of curl. The warning mentioned in this issue only appears if curl is already installed and up-to-date on the Mac machine. Therefore, it’s a good idea to suppress this warning now. There is no difference in working with this change.

vaisakhkannan:
Reference link for workflow run with this Fix : https://github.com/OpenLiberty/liberty-tools-intellij/actions/runs/10414819163/job/28847425223?pr=908

Reference link for workflow run without Fix
https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/10406582845/job/28819943167

NOTE: Check the Step `Install required integration test software`

mrglavas (approved):
Looks good. Thanks!

aparnamichael (approved):
Looks good!!

PR #909, opened 2024-08-12
Examine the UI for IntelliJ indexing and wait

### LINKED ISSUE #886: testInsertJakartaCodeSnippetIntoJavaPart()FAILED due to NoSuchElementException
While running UI tests in windows, GradleSingleModJakartaLSTest > `testInsertJakartaCodeSnippetIntoJavaPart`() is getting failed quite a few times due to ` java.util.NoSuchElementException: List is empty.`. If we re ran the tests a couple of times it is getting success.

Attaching failure logs - https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/10064285765/job/27842019395#step:10:836

turkeylurkey:
There is no issue for this fix but it affects several of the issues we have open for automated testing. For instance Anusree was looking at an issue on Windows and she noticed this problem.

Fixes #886

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
@@ -316,7 +316,7 @@ public abstract class SingleModMPLSTestCommon {
 
         // pre-open project tree before attempting to open files needed by testcases
         ProjectFrameFixture projectFrame = remoteRobot.find(ProjectFrameFixture.class, Duration.ofMinutes(2));
-        JTreeFixture projTree = projectFrame.getProjectViewJTree(projectName);
+        UIBotTestUtils.waitForIndexingToStop(remoteRobot, 1200); // Indexing on a slow VM
 
         UIBotTestUtils.openFile(remoteRobot, projectName, "ServiceLiveHealthCheck", projectName, "src", "main", "java", "io.openliberty.mp.sample", "health");
         UIBotTestUtils.openFile(remoteRobot, projectName, "microprofile-config.properties", projectName, "src", "main", "resources", "META-INF");
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
@@ -316,6 +316,7 @@ public abstract class SingleModMPLSTestCommon {
 
         // pre-open project tree before attempting to open files needed by testcases
         ProjectFrameFixture projectFrame = remoteRobot.find(ProjectFrameFixture.class, Duration.ofMinutes(2));
+        JTreeFixture projTree = projectFrame.getProjectViewJTree(projectName);
         UIBotTestUtils.waitForIndexingToStop(remoteRobot, 1200); // Indexing on a slow VM
 
         UIBotTestUtils.openFile(remoteRobot, projectName, "ServiceLiveHealthCheck", projectName, "src", "main", "java", "io.openliberty.mp.sample", "health");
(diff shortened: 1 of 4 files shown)


turkeylurkey:
@mezarin , @ajm01  I added you guys as optional reviewers in case you have valuable guidance. Thank you!

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
```diff
@@ -316,7 +316,7 @@ public static void prepareEnv(String projectPath, String projectName) {
 
         // pre-open project tree before attempting to open files needed by testcases
         ProjectFrameFixture projectFrame = remoteRobot.find(ProjectFrameFixture.class, Duration.ofMinutes(2));
-        JTreeFixture projTree = projectFrame.getProjectViewJTree(projectName);
```
mrglavas, line 319: The comment suggests that this call to `getProjectViewTree()` was intentional and Andy/Ed might have thought it was necessary to have here. Did you mean to remove this? They would more likely have a better understanding of why this was here, assuming they remember.

turkeylurkey: I only removed it because the result was not used but it might be there for side effects. Unfortunately I wrote this some months ago so I will need to redo the investigation.

turkeylurkey: My reason was that projTree was not used but now projectFrame is not used. Restoring old code.

src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
```diff
@@ -1761,6 +1761,22 @@ public static void waitForLTWTextAreaMessage(RemoteRobot remoteRobot, String mes
...
+        String xPath = "//div[@class='InlineProgressPanel']";
+        Locator progressPanelLocator = byXpath(xPath);
+        JLabelFixture progressPanelFixture = projectFrame.find(JLabelFixture.class, progressPanelLocator, Duration.ofSeconds(10));
+
+        RepeatUtilsKt.waitFor(Duration.ofSeconds(waitTime),
+                Duration.ofSeconds(1),
+                "Waiting for indexing message to disappear e.g. Indexing Java 17...",
+                "Indexing did not appear in the Liberty tool window",
+                () -> progressPanelFixture.findAllText().isEmpty());
+    }
```
TrevCraw, line 1779: Is it possible to hit a scenario where we are checking for indexing before indexing has started? If so, how can we handle that?

mrglavas: I think your question highlights just how tricky it is to make these UI tests reliable. The `RemoteRobot` is provided by JetBrains, right? I wonder if they have a guide or best practices on how to use it so that you get deterministic results when you apply it to the IDE.

turkeylurkey: Ed or Andy faced this issue when detecting the "empty liberty tools window" test. They waited 2s before checking for indexing before checking the window. I'll try to do the same.

turkeylurkey: The test for the empty liberty tools window actually waits for indexing to start and then waits for it to stop. In the current case we could do a similar thing, check the latest code. Let me know if it has become too complex.

mrglavas (approved):
Reviewed changes with Paul. Looks good. Thanks!

turkeylurkey:
Mac test fails with this error #913 
image

aparnamichael (approved):
Looks good!!

PR #917, opened 2024-08-16
Add a 5s delay after clicking on a quick fix

### LINKED ISSUE #916: Automated UI test too quick to validate change to file after using quick fix
In this screen shot we see that cursor just before it clicks the quick fix.
image
After the click the fix is not applied immediately.
image
But after a few seconds the fix appears. Yet the test fails. 

The validation code is extremely quick to run. In this case the test reads the source file and looks for the fix string. Most likely we just need to wait a certain time before checking for the correct code.

turkeylurkey:
Fixes #916

This affects four test cases that have quick fixes.


mrglavas (approved):
Looks good.

aparnamichael (approved):
Looks good!

PR #930, opened 2024-08-27
Document Recording all test videos

### LINKED ISSUE #928: Document Recording all test videos
Currently, we only record videos of failed tests. However, In some case if we want to review what occurred in the test prior to the failure, it would be beneficial to record videos for all tests.

anusreelakshmi934:
Fixes #928


anusreelakshmi934:
Raised to Main branch

Issue #895, opened 2024-08-02
Start running the cron job in the LTI repo 


anusreelakshmi934:
- [x] Add the necessary Webhook URLs and Secrets to the LTI Repo.
- [x] Remove `lsp4ij-market-0.0.2-integration` from matrix and add existing LTI Tags/Branches before merge to main
- [x] Merge the feature branch into the main
- [x] Verify cron jobs are running from main

vaisakhkannan:
We have added 3 secrets in LTI Repo 

1. SLACK_WEBHOOK_URL

--> This will enable us to receive notifications for `List of LSP4IJ PRs` that currently `opened` and `closed` recently in Slack channel.

2. WORKFLOW_BUILDER_WEBHOOK

--> This will enable us to respond to `LSP4IJ PRs` when `at least one PR` is present to display in slack channel. This PR may come from the `List of Open/New/Updated Pull Requests`, or from the `List of Pull Requests Closed Since the Last Cron Job`, which are identified using a Cron Job.

3. NO_PR_WORKFLOW_BUILDER_WEBHOOK

--> This will enable us to trigger a Slack notification when there are `no PRs to display`. This means both the `List of open/new/updated pull requests` and the `List of pull requests closed since the last cron job` are empty, as identified by a cron job.

aparnamichael:
Completed merging feature branch to main branch and cron job running successfully.

PR #932, opened 2024-08-28
Update existing LTI Tag as main branch 

vaisakhkannan:
Part of #895 

Updated existing LTI Tag as `main`. 
Since there is no existing LTI tag that are currently using LSP4IJ plugin, we specifying just LTI `main` branch only.


aparnamichael (approved):
Looks good!!

TrevCraw (approved):

PR #933, opened 2024-08-28
Update lsp4mp4ij component to align with Quarkus Tools for IntelliJ release 2.0.0 & Update LSP4MP to 0.12.0

### LINKED ISSUE #841: LTI 24.0.9: Align LSP4iJ, LSP4MP4iJ and LSP4MP for the release.
Current versions of the components on or being consumed by the LTI development stream:

**LSP4iJ**: 0.0.2
**LSP4MP4iJ**: snapshot taken on an arbitrary day from the Quarkus Tools main branch to match up with LSP4iJ 0.0.2
**LSP4MP**: 0.12.0-SNAPSHOT

============

Regarding LSP4iJ, there are a couple open issues we've been tracking:
https://github.com/redhat-developer/lsp4ij/issues/294
https://github.com/redhat-developer/lsp4ij/issues/324

We would prefer to resolve https://github.com/redhat-developer/lsp4ij/issues/294 before releasing and through testing we may identify additional issues that would impact the LTI release. This would require moving up to a future release of LSP4iJ (0.3.0?, 0.3.1?) which contains the fixes we require. That is the ideal outcome. We should get an understanding of what is likely to be available from LSP4iJ in the near future and assess whether that will be adequate from an LTI release.

============

Regarding LSP4MP4iJ, we should ideally be basing our snapshot off a release of Quarkus Tools so that it is easy to map this component back to its original source and also represents a version that Red Hat published to the Marketplace. We should check with the Quarkus Tools team on whether there is a release coming soon that we could align with.

============

Regrading LSP4MP, we prefer to have a dependency on a released version of LSP4MP. Currently we're relying on a snapshot of 0.12.0. We should ask the community whether the 0.12.0 is expected soon. The most recent release at the time of this writing appears to be: https://github.com/eclipse/lsp4mp/releases/tag/0.11.3. If 0.12.0 will not be available soon, we should determine if we can align with LSP4MP 0.11.3. Ideally we would align with LSP4MP 0.12.0 because LSP4MP4iJ is currently pulling in a snapshot of this release.

============

What we actually end up doing here depends very much on the activity in each of these dependencies. I've mostly focused on the ideal outcomes but we may need to consider alternatives in order to move forward.

============
Work items:

- [x] https://github.com/OpenLiberty/liberty-tools-intellij/issues/894
- [x] https://github.com/OpenLiberty/liberty-tools-intellij/issues/914
- [x] Update lsp4mp4ij component to align with Quarkus Tools for IntelliJ release 2.0.0: https://github.com/redhat-developer/intellij-quarkus/releases/tag/2.0.0
- [x] Update LSP4MP to 0.12.0: https://github.com/eclipse/lsp4mp/releases/tag/0.12.0

aparnamichael:
Fixes #841 
- Update lsp4mp4ij component to align with Quarkus Tools for IntelliJ release 2.0.0: https://github.com/redhat-developer/intellij-quarkus/releases/tag/2.0.0
-  Update LSP4MP to 0.12.0: https://github.com/eclipse/lsp4mp/releases/tag/0.12.0

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -67,7 +67,7 @@ configurations {
 
 dependencies {
 
-    implementation ('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.12.0-SNAPSHOT') {
+    implementation ('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.12.0') {
         exclude group: 'org.eclipse.lsp4j'
         exclude group: 'com.google.code.gson'
     }
--- a/build.gradle
+++ b/build.gradle
@@ -111,7 +111,7 @@ dependencies {
     testImplementation 'com.automation-remarks:video-recorder-junit5:2.0'
 
     // define jars to grab locally (if falling back to mavenLocal() repo)
-    lsp('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.12.0-SNAPSHOT:uber') {
+    lsp('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.12.0:uber') {
         transitive = false
     }
     lsp('org.eclipse.lemminx:org.eclipse.lemminx:0.26.1:uber') {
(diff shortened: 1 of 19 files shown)


build.gradle
```diff
@@ -67,7 +67,7 @@ configurations {
 
 dependencies {
 
-    implementation ('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.12.0-SNAPSHOT') {
+    implementation ('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.12.0') {
```
TrevCraw, line 70: There is a second spot where LSP4MP needs to be updated: https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/build.gradle#L109

TrevCraw (changes requested):

turkeylurkey (approved):
Looks good, thank you.

build.gradle
aparnamichael: Done

TrevCraw (approved):

PR #970, opened 2024-09-12
issue 955 - clear run in container module setting on run config deletion

### LINKED ISSUE #955: Deletion of configuration does not remove 'Start in Container' Setting when enabled
Steps to Reproduce 

1. Create a new Liberty run/debug config through the “Run” → “Edit configurations” window in IntelliJ
2. Select the “Run in container” checkbox and add a dev mode parameter (ie. -DhotTests)

Unknown-13

3. Run this configuration
4. Ensure devc starts up and the added parameter is part of the devc command in the terminal
5. Stop the dev mode
6. Click Run > Edit Configurations. In the Run/Debug Configurations window select the run config created and click on ‘-’ to delete it
7. Apply and then click OK ( Run button won't work here)
8.  run the Liberty “Start” action

I can still see the `Start in container` action performing, but the added `Start parameter` has been removed. This issue needs to be fixed. When I checked this feature in the LTI `24.0.6` release, I observed the same behaviour. I believe this could be considered an enhancement/Bug, as deleting a configuration should remove all associated settings that were made in that configuration.

aparnamichael:
Fixes #955 - Deletion of configuration does not remove 'Start in Container' Setting when enabled

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
+++ b/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
@@ -42,7 +42,7 @@ public class LibertyRunManagerListener implements RunManagerListener {
                 VirtualFile vBuildFile = VfsUtil.findFile(Paths.get(runConfig.getBuildFile()), true);
                 LibertyModule libertyModule = libertyModules.getLibertyModule(vBuildFile);
                 if (libertyModule != null) {
-                    libertyModule.setRunInContainer(false); // clear the run in configuration checkbox
+                    libertyModule.setRunInContainer(false); // clear the run in container checkbox
                     if (libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
                         libertyModule.setCustomStartParams(""); // clear the custom params
                     }
--- a/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
+++ b/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
@@ -43,6 +43,7 @@ public class LibertyRunManagerListener implements RunManagerListener {
                 if (libertyModule != null && libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
                     libertyModule.setCustomStartParams(""); // clear the custom params
                 }
+                libertyModule.setRunInContainer(false);
             } catch (Exception e) {
                 LOGGER.warn(String.format("Unable to clear custom start parameters for Liberty run configuration associated with: %s. Could not resolve build file.", runConfig.getBuildFile()), e);
             }
--- a/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
+++ b/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
@@ -28,7 +28,8 @@ public class LibertyRunManagerListener implements RunManagerListener {
     protected static Logger LOGGER = Logger.getInstance(LibertyRunManagerListener.class);
 
     /**
-     * When a Liberty run configuration is removed, clear custom start parameters from Liberty module
+     * When a Liberty run configuration is removed, clear custom start parameters and run in container
+     * from Liberty module
      *
      * @param settings
      */
--- a/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
+++ b/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
@@ -41,11 +41,9 @@ public class LibertyRunManagerListener implements RunManagerListener {
             try {
                 VirtualFile vBuildFile = VfsUtil.findFile(Paths.get(runConfig.getBuildFile()), true);
                 LibertyModule libertyModule = libertyModules.getLibertyModule(vBuildFile);
-                if (libertyModule != null) {
-                    libertyModule.setRunInContainer(false); // clear the run in container setting for the Liberty module.
-                    if (libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
+                if (libertyModule != null && libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
+                        libertyModule.setRunInContainer(false); // clear the run in container setting for the Liberty module.
                         libertyModule.setCustomStartParams(""); // clear the custom params
-                    }
                 }
             } catch (Exception e) {
                 LOGGER.warn(String.format("Unable to clear custom start parameters for Liberty run configuration associated with: %s. Could not resolve build file.", runConfig.getBuildFile()), e);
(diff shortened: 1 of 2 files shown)


turkeylurkey:
In LibertyModule it says
```
// FIXME not currently being used, need to enable runInContainer checkbox in LibertyRunConfiguration see https://github.com/OpenLiberty/liberty-tools-intellij/issues/160
    private boolean runInContainer;
```
Could you take this comment out since it is no longer applicable?

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
```diff
@@ -40,8 +41,11 @@ public void runConfigurationRemoved(@NotNull RunnerAndConfigurationSettings sett
             try {
                 VirtualFile vBuildFile = VfsUtil.findFile(Paths.get(runConfig.getBuildFile()), true);
                 LibertyModule libertyModule = libertyModules.getLibertyModule(vBuildFile);
-                if (libertyModule != null && libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
-                    libertyModule.setCustomStartParams(""); // clear the custom params
+                if (libertyModule != null) {
+                    libertyModule.setRunInContainer(false); // clear the run in container checkbox
```
turkeylurkey, line 45: Here we can see that every time `setCustomStartParams()` is called we also need to call `setRunInContainer()`. This is done in `LibertyRunConfiguration`. Could you open an issue to redesign this interface so that the params and the boolean must be set in one call? Now may not be a good time to redesign this interface.

turkeylurkey (changes requested):

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
```diff
@@ -40,8 +41,11 @@ public void runConfigurationRemoved(@NotNull RunnerAndConfigurationSettings sett
...
+                    if (libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
```
TrevCraw, line 46: @aparnamichael do you know the reasoning behind this check? What is the reason we don't just go ahead and clear the custom params every time a run configuration is removed?

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
TrevCraw, line 45: I think the comment "clear the run in container checkbox" could be confusing, since it is not actually doing anything to the run in container checkbox directly. This is setting the libertyModule to use a regular start instead of a container start.

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
mrglavas: @aparnamichael I was just reviewing with @TrevCraw and have similar questions.

aparnamichael: @mrglavas @TrevCraw These condition is there in the initial file commit. I also think that condition is not needed. But not sure whether that will introduce any new bugs.

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
aparnamichael: @turkeylurkey This means we should have a single method for eg: setParamsAndRunInContainer(String params, Boolean runInContainer)  in LibertyRunConfiguration ?
If we implement a single method like this, when a user checks the 'Run In Container' option, we also need to pass parameters along with the action of enabling or disabling the checkbox, correct?

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
TrevCraw: Looking into this further, it looks like the condition is to ensure that the run config being deleted is the one that was last run for this particular Liberty module. One Liberty module can have many run configs associated with it. If a user deletes a run config that was not the last one run, then we should not be resetting the Liberty module params to the default values. The same logic should apply for the "run in container" setting.

TrevCraw: As a side note, the current "check" to see if a config is last one run is not foolproof. It is possible configs could be identical. Configs could also have all the same custom dev mode params and have other differences (like the "run in container" setting). 

@mrglavas I think this just further shows a larger re-design is required, as you recommended in our discussion. 

@aparnamichael, as @turkeylurkey suggested in another comment, please open an issue for redesign. I would suggest a generic issue like "Redesign the relationship between Liberty module and run configurations". We can put the specific suggestion from @turkeylurkey in there, but @mrglavas recommended an even bigger overhaul of the design.

aparnamichael: @turkeylurkey and @TrevCraw created a new issue for re-design. https://github.com/OpenLiberty/liberty-tools-intellij/issues/977

turkeylurkey: Added more detail to that issue.

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
turkeylurkey: Yes, but we will have the Run configuration open in the UI when the checkbox is used. This is part of #977.

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
```diff
@@ -40,8 +41,11 @@ public void runConfigurationRemoved(@NotNull RunnerAndConfigurationSettings sett
             try {
                 VirtualFile vBuildFile = VfsUtil.findFile(Paths.get(runConfig.getBuildFile()), true);
                 LibertyModule libertyModule = libertyModules.getLibertyModule(vBuildFile);
-                if (libertyModule != null && libertyModule.getCustomStartParams().equals(runConfig.getParams())) {
-                    libertyModule.setCustomStartParams(""); // clear the custom params
+                if (libertyModule != null) {
+                    libertyModule.setRunInContainer(false); // clear the run in container setting for the Liberty module.
```
turkeylurkey, line 45: The change to `run in container` should only happen if the parameters are equal. Move this line after the `setCustomStartParams("")` line, inside the `if`.

turkeylurkey (changes requested):

TrevCraw (approved):

turkeylurkey (approved):
Thank you.

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunManagerListener.java
turkeylurkey: Looks good.

Issue #971, opened 2024-09-12
Mac automated UI test failure: initializationErrors


turkeylurkey:
Two test cases fail during initialization so there is no movie in the build output. These movies are used to debug the test failure so we'll need to debug these failures by adding debug logging.

- GradleSingleModMPLSTest > initializationError FAILED
- GradleSingleModMPProjectTest > initializationError FAILED

> GradleSingleModMPLSTest > initializationError FAILED
>     com.intellij.remoterobot.utils.WaitForConditionTimeoutException: Exceeded timeout (PT10M) for condition function (Failed to find 'ComponentFixture' by '//div[@class='Tree' and @name='LibertyTree' and contains(@visible_text, 'sampleGradleMPLSApp')]' in 600s) 
>         at com.intellij.remoterobot.utils.RepeatUtilsKt.waitFor(RepeatUtils.kt:51)
>         at com.intellij.remoterobot.utils.RepeatUtilsKt.waitFor$default(RepeatUtils.kt:37)
>         at com.intellij.remoterobot.SearchContext$find$1.invoke(SearchContext.kt:53)
>         at com.intellij.remoterobot.SearchContext$find$1.invoke(SearchContext.kt:51)
>         at com.intellij.remoterobot.stepsProcessing.StepWorkerKt.step(StepWorker.kt:23)
>         at com.intellij.remoterobot.SearchContext$DefaultImpls.find(SearchContext.kt:51)
>         at com.intellij.remoterobot.fixtures.ContainerFixture.find(ContainerFixture.kt:12)
>         at io.openliberty.tools.intellij.it.fixtures.ProjectFrameFixture.getTree(ProjectFrameFixture.java:215)
>         at io.openliberty.tools.intellij.it.UIBotTestUtils.validateImportedProjectShowsInLTW(UIBotTestUtils.java:382)
>         at io.openliberty.tools.intellij.it.SingleModMPLSTestCommon.prepareEnv(SingleModMPLSTestCommon.java:317)
>         at io.openliberty.tools.intellij.it.GradleSingleModMPLSTest.setup(GradleSingleModMPLSTest.java:45)

> GradleSingleModMPProjectTest > initializationError FAILED
>     com.intellij.remoterobot.utils.WaitForConditionTimeoutException: Exceeded timeout (PT2M) for condition function (Failed to find 'Welcome Frame' by 'FlatWelcomeFrame type' in 120s) 
[...]

Issue #945, opened 2024-09-05
Windows test, hover on text: NONE did not trigger a pop-up window


anusreelakshmi934:
While running the LTI build in windows we are coming across this issue.
In 
GradleSingleModLSTest > testDiagnosticInServerEnv() FAILED
    java.lang.RuntimeException: Hover on text: NONE did not trigger a pop-up window to open
Caused by:
        com.intellij.remoterobot.utils.WaitForConditionTimeoutException: Exceeded timeout (PT5S) for condition function (Failed to find 'Container' by '//div[@class='HeavyWeightWindow']//div[@class='JEditorPane']' in 5s) 
Its not getting the required text while hovering.

Attaching the build that ran - https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/10714632594/job/29708636124#step:10:901

<img width="580" alt="image" src="https://github.com/user-attachments/assets/0f1c947a-be6b-40d3-bb8d-a26336294562">

turkeylurkey:
Similar to the Mac issue #972

vaisakhkannan:
As part of the Windows test failures, the `testDiagnosticInServerEnv()` method is consistently failing in our current main code. In most builds, this is the only test that fails.

As a workaround, I tried different scenarios to check if the tests would pass:
1. Updated the value from "NONE" to "none".
        Out of 10 builds, builds 1, 2, 7, and 8 were successful.
        Build 3: 12 failures without the hover issue.
        Builds 4 and 9: 1 failure due to the hover issue.
        Build 5: 19 failures without the hover issue.
        Build 6: 10 failures without the hover issue.
        Build 10: 1 failure (without the hover issue) – MavenSingleModNLTRestProjectTest > initializationError FAILED.
    Summary: Only 2 builds failed due to the hover issue in server.env.
    Reference: https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/10989098813
    
2.Tried closing and reopening server.env after inserting the invalid value "NONE".
   Out of 10 builds:
      Builds 1, 3, 4, 6, 7, 9, and 10 passed.
      Build 2: 1 failure – testStartWithParamsActionUsingPopUpMenu() FAILED (no hover issue).
      Build 8: 2 failures – testInsertLibertyConfigIntoServerEnv() FAILED, testMPQuickFixInJavaFile() FAILED.
      Build 5: In progress.
    Summary: There is no hover issue in server.env on the diagnostic.
[...]

PR #980, opened 2024-09-17
Check test reports when Maven report generator 3.4 and 3.5 is used

### LINKED ISSUE #940: Add new tests to cover the two locations for test reports
Refer to https://github.com/OpenLiberty/liberty-tools-intellij/issues/939

Add automated tests to cover the following three scenarios:
1. Test report in new location
2. Test report in old location
3. Test report in both locations

turkeylurkey:
Updated two test cases to focus on the report generator. 

Existing test case `MavenSingleModMPProjectTest` already tests the integration and unit test reports. We delete the Maven cache of the Maven plugins so that they must be downloaded when used. The report generator plugin is not called out in the pom so when we run the tests dev mode downloads the latest report generator, 3.5.0. The test also deletes existing reports before it begins so we validate that the 3.5 test report is generated and the 3.4 is not.

Existing test case `MavenSingleModMPSIDProjectTest` also tests the reports. Again we delete the Maven cache and the existing test reports. This test case copies all the files into a new directory so we can safely modify pom.xml to specify Maven report generator 3.4.0 (this will all be deleted afterwards). When tests are run we validate that the 3.4 location was used and the 3.5 location is not.

Fixes #940

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
@@ -200,6 +200,6 @@ public class GradleSingleModMPProjectTest extends SingleModMPProjectTestCommon {
     @Override
     public void validateTestReportsExist() {
         //TODO: rewrite validateTestReportExists() to accept one argument or to accept a null as the second argument
-        TestUtils.validateTestReportExists(TEST_REPORT_PATH, TEST_REPORT_PATH);
+        TestUtils.validateTestReportExists(TEST_REPORT_PATH.toFile());
     }
 }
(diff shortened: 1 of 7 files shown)


turkeylurkey:
Windows: `148 tests completed, 1 failed, 16 skipped` The one failure is `Hover on text: NONE did not trigger a pop-up window`, a known failure #945

turkeylurkey:
Mac: `123 tests completed, 3 failed, 12 skipped` All three are initialization errors so no videos available. Similar to #971

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
```diff
@@ -200,6 +200,6 @@ public void deleteTestReports() {
     @Override
     public void validateTestReportsExist() {
         //TODO: rewrite validateTestReportExists() to accept one argument or to accept a null as the second argument
```
aparnamichael, line 202: Is this TODO statement still needed?

Issue #1024, opened 2024-10-16
Reduce code duplication in PersistenceMapKeyDiagnosticsCollector.


mrglavas:
The `collectDiagnostics()` method contains two loops over the methods and fields of a class. The block of code within each of those loops is identical. This should either be collapsed into one loop or refactor the block into its own method that each of the loops call.

This code duplication came from the port from LSP4Jakarta (https://github.com/eclipse/lsp4jakarta/blob/main/jakarta.jdt/org.eclipse.lsp4jakarta.jdt.core/src/main/java/org/eclipse/lsp4jakarta/jdt/internal/persistence/PersistenceMapKeyDiagnosticsParticipant.java). This improvement should be fed back to that project.

PR #1037, opened 2024-10-18
PersistenceMapKeyDiagnosticsCollector: Move body of the for loops over methods and fields into a common method.

mrglavas:
Resolves https://github.com/OpenLiberty/liberty-tools-intellij/issues/1024

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
@@ -18,16 +18,12 @@ import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.AbstractDiagnosticsColle
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.Messages;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 
 public class PersistenceMapKeyDiagnosticsCollector extends AbstractDiagnosticsCollector {
-
-    private static final Logger log = LoggerFactory.getLogger(PersistenceMapKeyDiagnosticsCollector.class);
 
     public PersistenceMapKeyDiagnosticsCollector() {
         super();
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2020, 2023 IBM Corporation, Ankush Sharma and others.
+ * Copyright (c) 2020, 2024 IBM Corporation, Ankush Sharma and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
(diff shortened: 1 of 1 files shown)


turkeylurkey (approved):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
```diff
@@ -37,81 +37,56 @@ protected String getDiagnosticSource() {
...
-                        validateMapKeyJoinColumnAnnotations(mapKeyJoinCols, field, unit, diagnostics);
-                    }
+                    collectDiagnostics(unit, diagnostics, type, field);
+                }
+            }
+        }
+    }
+
+    private void collectDiagnostics(PsiJavaFile unit, List<Diagnostic> diagnostics,
+                                    PsiClass type, PsiJvmModifiersOwner fieldOrProperty) {
```
TrevCraw, line 60: Looking at the two spots where this method is called, would `fieldOrMethod` make more sense for the last param name than `fieldOrProperty`?

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
```diff
@@ -37,81 +37,56 @@ protected String getDiagnosticSource() {
...
+                    hasMapKeyAnnotation = true;
+                else if (PersistenceConstants.MAPKEYCLASS.equals(matchedAnnotation))
+                    hasMapKeyClassAnnotation = true;
+                else if (PersistenceConstants.MAPKEYJOINCOLUMN.equals(matchedAnnotation)) {
+                    mapKeyJoinCols.add(annotation);
                 }
             }
         }
+        if (hasMapKeyAnnotation && hasMapKeyClassAnnotation) {
+            // A single field or property cannot have the same
```
TrevCraw, line 79: I know this comment did not originate with this PR, but I'm not clear on what it means.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
```diff
@@ -37,81 +37,56 @@ protected String getDiagnosticSource() {
...
+            diagnostics.add(createDiagnostic(fieldOrProperty, unit,
+                    Messages.getMessage("MapKeyAnnotationsNotOnSameField"),
+                    PersistenceConstants.DIAGNOSTIC_CODE_INVALID_ANNOTATION, null,
+                    DiagnosticSeverity.Error));
+        }
+        // If we have multiple MapKeyJoinColumn annotations on a single method or property we must
```
TrevCraw, line 85: Just noting that the `fieldOrProperty` param name seems off based on this comment.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
mrglavas: It's funny. I read this too and wanted to improve the comment, but also couldn't make sense of what the original author meant. I think this just came along for the ride from the port from LSP4Jakarta so would expect to see this comment in that code base too. I opted to just leave it alone.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
mrglavas: It's the correct terminology. See the Jarkata Persistence specification.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
mrglavas: See my response to the other comment. It's the correct terminology for the type of method. Entities can use either field or property based access.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceMapKeyDiagnosticsCollector.java
mrglavas: I did notice typo. it should say "... on a single field or property ..." instead of "... on a single method or property ...". I'll fix that.

turkeylurkey (approved):

TrevCraw (approved):
Looks good, thanks @mrglavas.

Please also make sure to open the issues we discussed:
1. Are we looping through a correct set of methods?
2. Are the constraints being checked here correct?

mrglavas:
> Looks good, thanks @mrglavas.
> 
> Please also make sure to open the issues we discussed:
> 
>     1. Are we looping through a correct set of methods?
> 
>     2. Are the constraints being checked here correct?

Yes, on my TODO list to open against https://github.com/eclipse/lsp4jakarta.

Issue #767, opened 2024-05-22
IntelliJ 2024.2: ProcessCanceledException class hierarchy changed. Resolve issues that arise from this API change.


mrglavas:
Change in IntelliJ 2024.2 https://github.com/JetBrains/intellij-community/commit/12e78f18d6352502c7cc942dc677f02bb9b8b257#diff-53953a080bfbb7f3d529336a7318cfde6b06098f5fbb136f7c58c3fad831f500
See resolution in LSP4iJ: https://github.com/redhat-developer/lsp4ij/pull/293.

LTI has multiple places with multi catch of `CancellationException` and `ProcessCanceledException` and will face the same problem.

LSP4iJ compilation errors that occurred with IntelliJ 2024.2:
![image](https://github.com/OpenLiberty/liberty-tools-intellij/assets/890521/409ace29-bf66-43a1-b252-c113cd2d43e1)

vaisakhkannan:
Usages of the following
catch (IndexNotReadyException | ProcessCanceledException | CancellationException e)

- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.getScopeAnnotations()
- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/annotations/PostConstructReturnTypeQuickFix.resolveCodeAction()
- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.resolveRemoveConstraintAnnotationsCodeAction()
- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.resolveStaticModifierCodeAction()
- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanNoArgConstructorQuickFix.resolveCodeAction()
- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanQuickFix.resolveCodeAction()
- [ ] src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/codeAction/proposal/quickfix/InsertAnnotationMissingQuickFix.resolveCodeAction()
[...]

PR #1047, opened 2024-10-22
Fix usage of ProcessCanceledException to support 2024.2

vaisakhkannan:
Part of #767 

I have created a new common class called `ExceptionUtil`, which other quick-fix classes use to call the common exception methods via `lambda expression`, in order to avoid code duplication.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
@@ -22,6 +22,7 @@ import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.stream.Collectors;
 import java.util.Collections;
+import java.util.function.Supplier;
 
 /**
  * Returns the list of recognised defining annotations applied to a
@@ -33,19 +34,28 @@ import java.util.Collections;
  */
 public class AnnotationUtil {
     public static List<String> getScopeAnnotations(PsiClass type, Set<String> scopes) {
+        return executeWithExceptionHandling(() ->
+                        // Construct a stream of only the annotations applied to the type that are also
+                        // recognised annotations found in scopes.
+                        Arrays.stream(type.getAnnotations())
+                                .map(annotation -> annotation.getNameReferenceElement().getQualifiedName())
+                                .filter(scopes::contains)
+                                .distinct()
+                                .collect(Collectors.toList()),
+                Collections::emptyList
+        );
+    }
+
+    private static <T> T executeWithExceptionHandling(Supplier<T> action, Supplier<T> fallback) {
         try {
-            // Construct a stream of only the annotations applied to the type that are also
-            // recognised annotations found in scopes.
-            return Arrays.stream(type.getAnnotations()).map(annotation -> annotation.getNameReferenceElement().getQualifiedName())
-                    .filter(scopes::contains).distinct().collect(Collectors.toList());
+            return action.get();
         } catch (ProcessCanceledException e) {
-            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multicatch to keep backward compatibility
-            //TODO delete block when minimum required version is 2024.2
+            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multi-catch to keep backward compatibility
             throw e;
         } catch (IndexNotReadyException | CancellationException e) {
             throw e;
         } catch (Exception e) {
-            return Collections.<String>emptyList();
+            return fallback.get(); // Return fallback value
         }
     }
 }
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
@@ -40,8 +40,10 @@ public class AnnotationUtil {
                         .filter(scopes::contains)
                         .distinct()
                         .collect(Collectors.toList()),
-                Collections::emptyList,  // Fallback value in case of exception
-                e -> LOGGER.log(Level.WARNING, "Error while calling getScopeAnnotations", e)
+                e -> {
+                    LOGGER.log(Level.WARNING, "Error while calling getScopeAnnotations", e);
+                    return Collections.<String>emptyList();
+                }
         );
     }
 }
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
@@ -12,17 +12,16 @@
 
 package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij;
 
-import com.intellij.openapi.progress.ProcessCanceledException;
-import com.intellij.openapi.project.IndexNotReadyException;
 import com.intellij.psi.PsiClass;
+import io.openliberty.tools.intellij.util.ExceptionUtil;
 
 import java.util.Arrays;
 import java.util.List;
 import java.util.Set;
-import java.util.concurrent.CancellationException;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import java.util.stream.Collectors;
 import java.util.Collections;
-import java.util.function.Supplier;
 
 /**
  * Returns the list of recognised defining annotations applied to a
@@ -33,29 +32,16 @@ import java.util.function.Supplier;
  * @return list of recognised defining annotations applied to a class
  */
 public class AnnotationUtil {
+    private static final Logger LOGGER = Logger.getLogger(AnnotationUtil.class.getName());
     public static List<String> getScopeAnnotations(PsiClass type, Set<String> scopes) {
-        return executeWithExceptionHandling(() ->
-                        // Construct a stream of only the annotations applied to the type that are also
-                        // recognised annotations found in scopes.
-                        Arrays.stream(type.getAnnotations())
-                                .map(annotation -> annotation.getNameReferenceElement().getQualifiedName())
-                                .filter(scopes::contains)
-                                .distinct()
-                                .collect(Collectors.toList()),
-                Collections::emptyList
+        return ExceptionUtil.executeWithExceptionHandling(
+                () -> Arrays.stream(type.getAnnotations())
+                        .map(annotation -> annotation.getNameReferenceElement().getQualifiedName())
+                        .filter(scopes::contains)
+                        .distinct()
+                        .collect(Collectors.toList()),
+                Collections::emptyList,  // Fallback value in case of exception
+                e -> LOGGER.log(Level.WARNING, "Error while calling getScopeAnnotations", e)
         );
-    }
-
-    private static <T> T executeWithExceptionHandling(Supplier<T> action, Supplier<T> fallback) {
-        try {
-            return action.get();
-        } catch (ProcessCanceledException e) {
-            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multi-catch to keep backward compatibility
-            throw e;
-        } catch (IndexNotReadyException | CancellationException e) {
-            throw e;
-        } catch (Exception e) {
-            return fallback.get(); // Return fallback value
-        }
     }
 }
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
@@ -38,7 +38,11 @@ public class AnnotationUtil {
             // recognised annotations found in scopes.
             return Arrays.stream(type.getAnnotations()).map(annotation -> annotation.getNameReferenceElement().getQualifiedName())
                     .filter(scopes::contains).distinct().collect(Collectors.toList());
-        } catch (IndexNotReadyException | ProcessCanceledException | CancellationException e) {
+        } catch (ProcessCanceledException e) {
+            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multicatch to keep backward compatibility
+            //TODO delete block when minimum required version is 2024.2
+            throw e;
+        } catch (IndexNotReadyException | CancellationException e) {
             throw e;
         } catch (Exception e) {
             return Collections.<String>emptyList();
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
@@ -1,4 +1,4 @@
-/* Copyright (c) 2022 IBM Corporation, Lidia Ataupillco Ramos and others.
+/* Copyright (c) 2022, 2024 IBM Corporation, Lidia Ataupillco Ramos and others.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
@@ -35,6 +35,8 @@ public class AnnotationUtil {
     private static final Logger LOGGER = Logger.getLogger(AnnotationUtil.class.getName());
     public static List<String> getScopeAnnotations(PsiClass type, Set<String> scopes) {
         return ExceptionUtil.executeWithExceptionHandling(
+                // Construct a stream of only the annotations applied to the type that are also
+                // recognised annotations found in scopes.
                 () -> Arrays.stream(type.getAnnotations())
                         .map(annotation -> annotation.getNameReferenceElement().getQualifiedName())
                         .filter(scopes::contains)
--- a/src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/validators/annotations/AnnotationValidator.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/validators/annotations/AnnotationValidator.java
@@ -96,8 +96,10 @@ public class AnnotationValidator {
 					}
 					return rule.validate(value);
 				},
-				() -> null,  // Fallback value supplier in case of exception
-				e -> LOGGER.log(Level.WARNING, e.getLocalizedMessage(), e)
+				e -> {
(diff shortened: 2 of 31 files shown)


TrevCraw (commented):
A couple overview thoughts:

1. We should not be using `System.out.println` anywhere, we should be writing to a log file.
2. In the spots where `System.out.println` is being used (the "unsuccessful" cases), has it been considered how we should handle being "unsuccessful"? Is it OK to just log an error and move on? Is there other handling we need to do?
3. I still see patterns of code that are duplicated quite often, mainly this pattern here:
```
Boolean success = ExceptionUtil.executeWithExceptionHandling(
				() -> {
					WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
					toResolve.setEdit(we);
					return true;
				},
				e -> LOGGER.log(Level.WARNING, "Unable to create workspace edit for code action to insert missing annotation", e)
		);
		if (success == null || !success) {
			System.out.println("An error occurred during the code action resolution.");
		}
```
It should be possible to write a common method to cover this code pattern.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/validators/annotations/AnnotationValidator.java
```diff
@@ -93,7 +93,10 @@ public String validate(String value, AnnotationAttributeRule rule) {
 				return null;
 			}
 			return rule.validate(value);
```
vaisakhkannan, line 95: For the code below, the usage of the try block is different, so I used standard exception handling for those cases.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
```diff
@@ -33,15 +34,28 @@
...
+        } catch (ProcessCanceledException e) {
+            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multi-catch to keep backward compatibility
+            throw e;
+        } catch (IndexNotReadyException | CancellationException e) {
             throw e;
         } catch (Exception e) {
-            return Collections.<String>emptyList();
+            return fallback.get(); // Return fallback value
         }
     }
```
mrglavas, line 60: I see this method is private and only called by `getScopeAnnotations()`. Is there some way to combine this with the more widely used utility method you introduced in `ExceptionUtil`? Otherwise this has just introduced more complexity without reducing code duplication in this spot.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/validators/annotations/AnnotationValidator.java
mrglavas: This looks structurally similar to what you refactored into the utility method. I'm not sure why the pattern wouldn't also apply here.

mrglavas:
> A couple overview thoughts:
> 
>     1. We should not be using `System.out.println` anywhere, we should be writing to a log file.
> 
>     2. In the spots where `System.out.println` is being used (the "unsuccessful" cases), has it been considered how we should handle being "unsuccessful"? Is it OK to just log an error and move on? Is there other handling we need to do?
> 
>     3. I still see patterns of code that are duplicated quite often, mainly this pattern here:
> 
> 
> ```
> Boolean success = ExceptionUtil.executeWithExceptionHandling(
> 				() -> {
> 					WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
> 					toResolve.setEdit(we);
> 					return true;
> 				},
> 				e -> LOGGER.log(Level.WARNING, "Unable to create workspace edit for code action to insert missing annotation", e)
> 		);
> 		if (success == null || !success) {
> 			System.out.println("An error occurred during the code action resolution.");
> 		}
> ```
> 
> It should be possible to write a common method to cover this code pattern.

I believe there's been some iteration on this PR since @TrevCraw 's initial review, but I'm still seeing duplication of the exception handling pattern in this PR. In the multiple place where the utility method is being used, I think those look great but wanted to point out that many of the changes made are in LSP4MP4IJ which is code that we basically copy out of Quarkus Tools. I think it would be beneficial to contribute a PR to IntelliJ Quarkus (https://github.com/redhat-developer/intellij-quarkus) which introduces the utility method which we would also use in Liberty Tools. This would eliminate the cost of having to apply these changes every time we do an update for LSP4MP4IJ.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/validators/annotations/AnnotationValidator.java
vaisakhkannan: Hi @mrglavas , I made the changes based on your suggestions

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/AnnotationUtil.java
vaisakhkannan: Yes @mrglavas , I made the changes , please take a look , Thanks

vaisakhkannan:
> > A couple overview thoughts:
> > ```
> > 1. We should not be using `System.out.println` anywhere, we should be writing to a log file.
> > 
> > 2. In the spots where `System.out.println` is being used (the "unsuccessful" cases), has it been considered how we should handle being "unsuccessful"? Is it OK to just log an error and move on? Is there other handling we need to do?
> > 
> > 3. I still see patterns of code that are duplicated quite often, mainly this pattern here:
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > Boolean success = ExceptionUtil.executeWithExceptionHandling(
> > 				() -> {
> > 					WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
> > 					toResolve.setEdit(we);
> > 					return true;
> > 				},
> > 				e -> LOGGER.log(Level.WARNING, "Unable to create workspace edit for code action to insert missing annotation", e)
> > 		);
> > 		if (success == null || !success) {
> > 			System.out.println("An error occurred during the code action resolution.");
> > 		}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It should be possible to write a common method to cover this code pattern.
> 
> I believe there's been some iteration on this PR since @TrevCraw 's initial review, but I'm still seeing duplication of the exception handling pattern in this PR. In the multiple place where the utility method is being used, I think those look great but wanted to point out that many of the changes made are in LSP4MP4IJ which is code that we basically copy out of Quarkus Tools. I think it would be beneficial to contribute a PR to IntelliJ Quarkus (https://github.com/redhat-developer/intellij-quarkus) which introduces the utility method which we would also use in Liberty Tools. This would eliminate the cost of having to apply these changes every time we do an update for LSP4MP4IJ.

Hi @mrglavas I can open PR against repo https://github.com/redhat-developer/intellij-quarkus to  update for `LSP4MP4IJ`

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/core/java/corrections/JavaDiagnosticsDefinition.java
```diff
@@ -50,49 +51,48 @@ public final class JavaDiagnosticsDefinition extends BaseKeyedLazyInstance<IJava
...
-            throw e;
-        } catch (Exception e) {
-            LOGGER.log(Level.WARNING, "Error while calling beginDiagnostics", e);
-        }
+        ExceptionUtil.executeWithExceptionHandling(
+            () -> {
+                getInstance().beginDiagnostics(context);
+                return true;
+            },
+                null,
```
mrglavas, line 68: `NullPointerException`? See my previous comment.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/core/java/corrections/JavaDiagnosticsDefinition.java
```diff
@@ -50,49 +51,48 @@ public final class JavaDiagnosticsDefinition extends BaseKeyedLazyInstance<IJava
...
-            throw e;
-        } catch (Exception e) {
-            LOGGER.log(Level.WARNING, "Error while calling endDiagnostics", e);
-        }
+        ExceptionUtil.executeWithExceptionHandling(
+            () -> {
+                getInstance().endDiagnostics(context);
+                return true;
+            },
+                null,
```
mrglavas, line 93: `NullPointerException`? See my previous comment.

src/main/java/io/openliberty/tools/intellij/util/ExceptionUtil.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public static void executeWithWorkspaceEditHandling(JavaCodeActionResolveContext context, ChangeCorrectionProposal proposal, CodeAction toResolve, Logger logger, String logMessage) {
+        executeWithExceptionHandling(
+            () -> {
+                WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
+                toResolve.setEdit(we);
+                return true;
+            },
+                null,
```
mrglavas, line 49: `NullPointerException`? See previous comments.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/core/java/validators/annotations/AnnotationValidator.java
```diff
@@ -88,17 +89,16 @@ public void unregisterRule(AnnotationRule rule) {
...
-			return null;
-		}
+		return ExceptionUtil.executeWithExceptionHandling(
+				() -> {
+					if (rule == null) {
+						return null;
+					}
+					return rule.validate(value);
+				},
+				null,  // Fallback value in case of exception
```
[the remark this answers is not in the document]
vaisakhkannan: @mrglavas , Thanks for the suggestions, Yes you are correct that will get a `NullPointerException`.
So I initially have a thought of changing code like below

```
public String validate(String value, AnnotationAttributeRule rule) {
    return ExceptionUtil.executeWithExceptionHandling(
        () -> {
            if (rule == null) {
                return null;
            }
            return rule.validate(value);
        },
        () -> null,  // Fallback value supplier in case of exception
        e -> LOGGER.log(Level.WARNING, e.getLocalizedMessage(), e)
    );
}
```

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/core/java/corrections/JavaDiagnosticsDefinition.java
vaisakhkannan: I made the change like this 

```
@Override
    public void beginDiagnostics(JavaDiagnosticsContext context) {
        ExceptionUtil.executeWithExceptionHandling(
            () -> {
                getInstance().beginDiagnostics(context);
                return true;
            },
            e -> {
                LOGGER.log(Level.WARNING, "Error while calling beginDiagnostics", e);
                return null;
            }
        );
    }
```

This won't give `NullPointerException`.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/core/java/corrections/JavaDiagnosticsDefinition.java
vaisakhkannan: Did change similar to above comment

```
@Override
    public void endDiagnostics(JavaDiagnosticsContext context) {
        ExceptionUtil.executeWithExceptionHandling(
            () -> {
                getInstance().endDiagnostics(context);
                return true;
            },
            e -> {
                LOGGER.log(Level.WARNING, "Error while calling endDiagnostics", e);
                return null;
            }
        );
    }
```

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
```diff
@@ -152,15 +151,16 @@ private void validateFallbackAnnotation(PsiMethod node, PsiAnnotation annotation
...
-		} catch (Exception e) {
-			throw e;
-		}
+		// Use lambda expression to avoid code duplication
+		String methodReturnTypeString = ExceptionUtil.executeWithExceptionHandling(
+			() -> {
+				PsiType methodReturnType = node.getReturnType();
+				return methodReturnType.getCanonicalText();
+			},
+				null,
```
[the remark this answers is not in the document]
vaisakhkannan: Thanks for the suggestion , I made the change in method `executeWithExceptionHandling`

please check my previous comment [here](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1047#discussion_r1828754097)

mrglavas:
> I am attaching the recording of the failed test and logging the error message.

@vaisakhkannan To be clear, I think what you're trying to show here is a demo of the fallback path and you've done that by forcing an exception to be thrown (with temporary code which divides by 0 that isn't in the PR). Thanks for showing that. I was concerned for a moment that this was about test failures in the build.

vaisakhkannan:
> done that by forcing an exception to be thrown (with temporary code which divides by 0 that isn't in the PR)

Yes @mrglavas , I tested reproducing exception by adding temporary code that forces an exception by dividing by zero, but that code isn't included in this PR.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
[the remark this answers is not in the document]
mrglavas: @vaisakhkannan It does not seem like you resolved this issue. It looks like the code is returning `null` instead of re-throwing the exception which is in the original code. If you're looking for ideas on how to handle this scenario, this post on Stack Overflow might be helpful: https://stackoverflow.com/questions/18198176/java-8-lambda-function-that-throws-exception

mrglavas (changes requested):
See my most recent comment.

src/main/java/io/openliberty/tools/intellij/util/ExceptionUtil.java
```diff
@@ -0,0 +1,76 @@
...
+    public static void executeWithWorkspaceEditHandling(JavaCodeActionResolveContext context, ChangeCorrectionProposal proposal, CodeAction toResolve, Logger logger, String logMessage) {
+        executeWithExceptionHandling(
+            () -> {
+                WorkspaceEdit we = context.convertToWorkspaceEdit(proposal);
+                toResolve.setEdit(we);
+                return true;
+            },
+            e -> {
+                logger.log(Level.WARNING, logMessage, e);
+                return false;
```
vaisakhkannan, line 72: @mrglavas ,I replaced `null` with `false` , so  it’s better to stick with `false` to clearly indicate failure without adding `null` ambiguity.

src/main/java/io/openliberty/tools/intellij/lsp4mp4ij/psi/internal/faulttolerance/java/MicroProfileFaultToleranceASTValidator.java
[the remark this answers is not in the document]
vaisakhkannan: @mrglavas, Thanks for review comment and link you shared, I modified the code to re-throwing the exception, We can't use `throw e`, so I used,
 `throw new RuntimeException("Failed to validate asynchronous annotation", e);` 
 
By throwing a RuntimeException, you're signaling that something went wrong.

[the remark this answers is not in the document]
mrglavas: @vaisakhkannan The link I shared demonstrates exactly how you can re-throw the exception using a custom functional interface, so this is technically possible to do. The code you are modifying comes from IntelliJ Quarkus and without in-depth understanding of all the paths we could enter this method, just changing this to a generic `RuntimeException` could change the behaviour. A wrapped exception is not the same. The goal here is to improve how we do this exception handling without changing what it's doing. If the current usage can only produce unchecked exceptions, perhaps we should change the catch from `Exception` to `RuntimeException` and then the function could re-throw the `RuntimeException` without any restrictions.

mrglavas (changes requested):
See my most recent comment about re-throwing exceptions.

src/main/java/io/openliberty/tools/intellij/util/ExceptionUtil.java
mrglavas: @vaisakhkannan Do all the callers understand the possible return values? What if the return type is Boolean? `false` might actually might be a normal return value. I know our component is the only caller of this utility method. I just want to make sure we understand the impact of this change.

src/main/java/io/openliberty/tools/intellij/util/ExceptionUtil.java
```diff
@@ -0,0 +1,76 @@
...
+            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multi-catch to keep backward compatibility
+            //TODO delete block when minimum required version is 2024.2
+            throw e;
+        } catch (IndexNotReadyException | CancellationException e) {
+            throw e;
+        } catch (Exception e) {
+            // Invoke the fallback function with the exception to generate a safe return value.
+            return fallback.apply(e);
+        }
+    }
```
mrglavas, line 52:
```suggestion
    public static <T> T executeWithExceptionHandling(Supplier<T> action, Function<RuntimeException, T> fallback) {
        try {
            return action.get();
        } catch (ProcessCanceledException e) {
            //Since 2024.2 ProcessCanceledException extends CancellationException so we can't use multi-catch to keep backward compatibility
            //TODO delete block when minimum required version is 2024.2
            throw e;
        } catch (IndexNotReadyException | CancellationException e) {
            throw e;
        } catch (RuntimeException e) {
            // Invoke the fallback function with the exception to generate a safe return value.
            return fallback.apply(e);
        }
    }
```

mrglavas: @vaisakhkannan As we discussed today, here's the suggested change to allow the fallback function to be able to re-throw exceptions.

vaisakhkannan: @mrglavas , I just saw your commit suggestion after I pushed the changes I made locally, so I didn’t receive the email at the same time. Could you please check the changes I pushed now?

vaisakhkannan: @mrglavas , I made the change

mrglavas (approved):
Looks good now. Thanks for addressing the review comments.

dessina-devasia (approved):
Looks good to me

PR #1051, opened 2024-10-23
Wait 5s after UI robot types in a config file

### LINKED ISSUE #945: Windows test, hover on text: NONE did not trigger a pop-up window

turkeylurkey:
The diagnostic fails to appear in the editor because the robot clicks on the editor tab *immediately* after typing the value (after the equals. sign). A bug or feature in IntelliJ makes it give up on the unrendered diagnostic when the user turns attention somewhere else.

Here is how I decided on the amount of time to wait. It may take 1s so I doubled it. On Mac it might be slow so I doubled it again and rounded up.

This fix relies on using the LCLS 2.2-SNAPSHOT. I do not include it here because someone else has been testing this and will make a separate pull request. Here is the test result when you do use 2.2-snapshot:
image

Fixes #945

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -1104,6 +1104,7 @@ public class UIBotTestUtils {
                 keyboard.enter();
 
                 keyboard.enterText(configNameSnippet);
+                TestUtils.sleepAndIgnoreException(5);
 
                 // Narrow down the config name completion suggestions in the pop-up window that is automatically
                 // opened as text is typed based on the value of configNameSnippet. Avoid hitting ctrl + space as it has the side effect of selecting
@@ -1123,6 +1124,7 @@ public class UIBotTestUtils {
                 keyboard.hotKey(VK_END);
 
                 keyboard.enterText(configValueSnippet);
+                TestUtils.sleepAndIgnoreException(5);
 
                 if (completeWithPopup) {
                     // Select the appropriate value completion suggestion in the pop-up window that is automatically


src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
```diff
@@ -1104,6 +1104,7 @@ public static void insertConfigIntoConfigFile(RemoteRobot remoteRobot, String fi
                 keyboard.enter();
 
                 keyboard.enterText(configNameSnippet);
+                TestUtils.sleepAndIgnoreException(5);
```
vaisakhkannan, line 1107: @turkeylurkey,  After checking the build results, I found that only one Windows build failed due to the `diagnostic not appearing in the server.env issue`. Apart from that, the other builds, including Mac and Linux, were successful. Therefore, the probability of failure due to the diagnostic not appearing in the `server.env` issue is very low. I believe this fix is good, and it should be fine to get green builds now, as the issue is not reproducible through manual testing.

TrevCraw:
Let's wait and see how things go with a fix for https://github.com/redhat-developer/lsp4ij/issues/577

src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
turkeylurkey: Thank you for the comment.

mrglavas (approved):
Looks good and nice to see the green Windows build. Thanks!

TrevCraw (commented):
Changes look good and green builds look even better. Can you please add some comments to the sleep lines to explain why they are needed?

TrevCraw:
To be clear, we believe this PR combined with this fix in LSP4IJ - https://github.com/redhat-developer/lsp4ij/pull/587 - will ultimately resolve the server.env test failure as reported in https://github.com/OpenLiberty/liberty-tools-intellij/issues/945.

turkeylurkey:
Check in tomorrow! 🕥

turkeylurkey:
10/31 Linux failure due to Liberty hung error.

mrglavas (approved):
Looks good. Thanks.

TrevCraw (approved):

PR #1056, opened 2024-10-29
Removed unused unqualified names from WebSocketConstants

### LINKED ISSUE #1031: websocket.WebSocketDiagnosticsCollector: Investigate and resolve issues with matching qualified names.
This class is within the `io.openliberty.tools.intellij.lsp4jakarta.lsp4ij` package.

See analysis and areas of concern on this issue: https://github.com/OpenLiberty/liberty-tools-intellij/issues/879.

anusreelakshmi934:
Fixes #1031

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
@@ -25,8 +25,7 @@ public class WebSocketConstants {
 
     public static final DiagnosticSeverity ERROR = DiagnosticSeverity.Error;
     public static final DiagnosticSeverity WARNING = DiagnosticSeverity.Warning;
-
-    public static final String DIAGNOSTIC_ERR_MSG = "Cannot calculate WebSocket diagnostics";
+    
 
     public static final String PATHPARAM_ANNOTATION = "jakarta.websocket.server.PathParam";
 
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
@@ -63,6 +63,14 @@ public class WebSocketConstants {
 
     public static final String[] WS_ANNOTATION_CLASS = { SERVER_ENDPOINT_ANNOTATION, CLIENT_ENDPOINT_ANNOTATION };
 
+    public static final String BOOLEAN = "java.lang.Boolean";
+    public static final String INTEGER = "java.lang.Integer";
+    public static final String LONG = "java.lang.Long";
+    public static final String DOUBLE = "java.lang.Double";
+    public static final String FLOAT = "java.lang.Float";
+    public static final String CHARACTER = "java.lang.Character";
+    public static final String BYTE = "java.lang.Byte";
+    public static final String SHORT = "java.lang.Short";
     public static final String STRING_CLASS_LONG = "java.lang.String";
     public static final String STRING_CLASS_SHORT = "String";
     public static final String READER_CLASS_LONG = "java.io.Reader";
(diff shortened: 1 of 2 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
```diff
@@ -102,9 +106,7 @@ public class WebSocketConstants {
             Arrays.asList("CloseReason", "Session"));
 
     public static final Set<String> RAW_WRAPPER_OBJS = new HashSet<>(
-            Arrays.asList("String", "Boolean", "Integer", "Long", "Double", "Float"));
-    public static final Set<String> WRAPPER_OBJS = RAW_WRAPPER_OBJS.stream().map(raw -> "java.lang.".concat(raw))
-            .collect(Collectors.toSet());
+            Arrays.asList(BOOLEAN, INTEGER, LONG, DOUBLE, FLOAT, CHARACTER, BYTE, SHORT, STRING_CLASS_LONG));
```
mrglavas, line 109: I think this needs some explanation since this changes the value of the constant. There are more names in this list than before. Are you sure that these additional classes belong here? Did you test this? I think even if this is an appropriate change to make the best thing to do process wise would have been to open a separate GitHub issue which describes the problem and address it in its own PR.

mrglavas (changes requested):
See review comments.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
turkeylurkey: These types came from the Jakarta specification so it would be best to check there.

anusreelakshmi934: @mrglavas, sorry the additional names were added by mistake; I’ll remove them. Aside from that, is this approach correct for addressing the issue?

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
```diff
@@ -58,12 +57,16 @@ public class WebSocketConstants {
...
 
     public static final String[] WS_ANNOTATION_CLASS = { SERVER_ENDPOINT_ANNOTATION, CLIENT_ENDPOINT_ANNOTATION };
 
+    public static final String BOOLEAN = "java.lang.Boolean";
+    public static final String INTEGER = "java.lang.Integer";
+    public static final String LONG = "java.lang.Long";
+    public static final String DOUBLE = "java.lang.Double";
+    public static final String FLOAT = "java.lang.Float";
     public static final String STRING_CLASS_LONG = "java.lang.String";
     public static final String STRING_CLASS_SHORT = "String";
```
mrglavas, line 71: I noticed that there are still several constants (such as `STRING_CLASS_SHORT` for unqualfied names that are still used by this class after your changes. I'm guessing they're being used somewhere but wondering whether you checked what is using them and why. If you haven't investigated this yet can you check? I suspect there might be another issue somewhere or at least obsolete code that we should clean up.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
mrglavas: @anusreelakshmi934 Yes, this is generally the right approach, though please see my other comment about investigating why some constants for unqualified names are still in use.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketConstants.java
anusreelakshmi934: I reviewed the usage of these names and identified that they were referenced [here](https://github.com/OpenLiberty/liberty-tools-intellij/blob/efe3fedbdff07739813239fb740b59b36a8efb75/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketDiagnosticsCollector.java#L391). According to [this commit](https://github.com/OpenLiberty/liberty-tools-intellij/commit/ca2f0c78fc21cc6e7769eb3d06bc5c2f72400a3c#diff-e1a14466e98b051537053ad0b6ba48a7e6e5fd0a2697f83cc3f70115ad461bf6L215-R189), `WebSocketConstants.SHORT_MESSAGE_CLASSES.contains(typeName)` was removed, leaving only `WebSocketConstants.LONG_MESSAGE_CLASSES.contains(typeName)` in use. Therefore, the switch case [here](https://github.com/OpenLiberty/liberty-tools-intellij/blob/efe3fedbdff07739813239fb740b59b36a8efb75/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/websocket/WebSocketDiagnosticsCollector.java#L391) became unnecessary. I have removed it and refactored the code accordingly.

mrglavas (approved):
Looks good. Thanks for following up and cleaning up the code.

Issue #1080, opened 2024-11-06
Jakarta: Create java.lang and java.math imposter classes for use in testing qualified names.


mrglavas:
To facilitate testing that the support for processing qualified names in the Jakarta diagnostics and quick fixes is correct, the test application source will need to reference classes which have similar names to the Java built-in types.

Write an imposter class (same simple name in a different package) for each of:

```
java.lang.Boolean
java.lang.Byte
java.lang.Character
java.lang.CharSequence
java.lang.Double
java.lang.Float
java.lang.Integer
java.lang.Long
java.lang.Short
java.lang.String

java.math.BigDecimal
java.math.BigInteger
```

PR #1081, opened 2024-11-06
Create java.lang and java.math imposter classes for use in testing qualified names.

mrglavas:
Resolves https://github.com/OpenLiberty/liberty-tools-intellij/issues/1080

Package mappings:
```
java.lang -> ng
java.math -> th
```

I explicitly chose these package names because they are suffixes of the Java built-in packages. Code which incorrectly matches qualified names using the `String.endsWith()` method can potentially be identified by using these imposter classes in the test application.


dessina-devasia (approved):
Looks good to me

turkeylurkey (approved):

PR #1083, opened 2024-11-06
Refactor the automated UI tests for easier maintenance

### LINKED ISSUE #1082: Refactor automated UI test classes to make them easier to maintain
The classes `SingleModMPProjectTestCommon` and `SingleModNLTRestProjectTestCommon` define a large number of abstract methods which are actually just getters. This means all the classes that extend these base classes must implement the methods in exactly the same way. It would be simpler if the base classes simply implemented the getters one time and the derived classes used them.

This requires the base classes to define local variables and to implement setters as well as getters but this is quite simple to do and hence worthwhile to allow this refactoring to go forward.

turkeylurkey:
The new copyright headers specify year 2023 which is the correct year.

Fixes #1082

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
@@ -17,15 +17,28 @@ import java.nio.file.Paths;
  * Tests Liberty Tools actions using a single module MicroProfile Gradle project.
  */
 public class GradleSingleModMPProjectTest extends SingleModMPProjectTestCommon {
+    /**
+     * Single module Microprofile project name.
+     */
+    private static final String SM_MP_PROJECT_NAME = "singleModGradleMP";
+
+    /**
+     * The path to the folder containing the test projects.
+     */
+    private static final String PROJECTS_PATH = Paths.get("src", "test", "resources", "projects", "gradle").toAbsolutePath().toString();
 
     /**
      * Prepares the environment for test execution.
      */
     @BeforeAll
     public static void setup() {
-        setSmMPProjectName("singleModGradleMP");
+        prepareEnv(PROJECTS_PATH, SM_MP_PROJECT_NAME);
+    }
+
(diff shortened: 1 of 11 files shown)


mrglavas:
I noticed this PR has moved away from an object oriented solution to state managed by static fields. When you were describing the refactoring I had imagined the state would be managed by an instance of the test class or some other object. Imagine a JUnit test runner that runs test cases in parallel. How would that work now? I don't think it would.

turkeylurkey:
@mrglavas Good points. I made the tests more OO.

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
```diff
@@ -88,131 +35,18 @@ public static void setup() {
...
-    }
-
-    /**
-     * Returns the build type category (either MAVEN_TYPE or GRADLE_TYPE)
-     */
-    @Override
-    public BuildType getBuildCategory() {
-        return BUILD_CATEGORY;
+    GradleSingleModMPProjectTest() {
+        setProjectsDirPath(PROJECTS_PATH);
```
anusreelakshmi934, line 39: It will be better to add some comments before each of these setting so that it makes it easy to quickly understand the purpose of each configuration ?

mrglavas (approved):
Looks good now. Thanks!

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
turkeylurkey: I was hoping these attributes were self explanatory and the comments in the base class would remove any ambiguity. Also, I put the project path and name first and other attributes later to indicate the importance of each.

anusreelakshmi934: Ok .Its fine then. Thanks!!@turkeylurkey @mrglavas.

anusreelakshmi934 (approved):
Looks Good to me

PR #1084, opened 2024-11-07
Display short names for Quick fixes - Issue#870

### LINKED ISSUE #870: Quick fix uses the fully qualified name in the display label
The BeanValidationQuickFix uses the fully qualified name in the display label. This can be hard to read, it should use the short name like we see in the diagnostics.

image

anusreelakshmi934:
Fixes #870

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.java
@@ -16,6 +16,7 @@ import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.project.IndexNotReadyException;
 import com.intellij.psi.*;
 import com.intellij.psi.util.PsiTreeUtil;
+import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.AbstractDiagnosticsCollector;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.JDTUtils;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.Messages;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction.proposal.ModifyModifiersProposal;
@@ -81,7 +82,7 @@ public class BeanValidationQuickFix implements IJavaCodeActionParticipant {
     private void removeConstraintAnnotationsCodeActions(Diagnostic diagnostic, JavaCodeActionContext context, List<CodeAction> codeActions) {
 
         final String annotationName = diagnostic.getData().toString().replace("\"", "");
-        final String name = Messages.getMessage("RemoveConstraintAnnotation", getSimpleName(annotationName));
+        final String name = Messages.getMessage("RemoveConstraintAnnotation", AbstractDiagnosticsCollector.getSimpleName(annotationName));
         Map<String, Object> extendedData = new HashMap<>();
         extendedData.put(ANNOTATION_NAME, annotationName);
         codeActions.add(JDTUtils.createCodeAction(context, diagnostic, name, getParticipantId(), extendedData));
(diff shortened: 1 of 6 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.java
```diff
@@ -81,7 +81,7 @@ public CodeAction resolveCodeAction(JavaCodeActionResolveContext context) {
     private void removeConstraintAnnotationsCodeActions(Diagnostic diagnostic, JavaCodeActionContext context, List<CodeAction> codeActions) {
 
         final String annotationName = diagnostic.getData().toString().replace("\"", "");
-        final String name = Messages.getMessage("RemoveConstraintAnnotation", annotationName);
+        final String name = Messages.getMessage("RemoveConstraintAnnotation", annotationName.substring(annotationName.lastIndexOf('.') + 1));
```
dessina-devasia, line 84: Rather than directly applying substring(), it will be better to get the simple name by using the function like below.

protected static String getSimpleName(String fqName) {
        int idx = fqName.lastIndexOf('.');
        if (idx != -1 && idx != fqName.length() - 1) {
            return fqName.substring(idx + 1);
        }
        return fqName;
    }

anusreelakshmi934: I have changed it accordingly. Thanks

dessina-devasia (approved):
Looks good for me

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.java
```diff
@@ -150,6 +150,17 @@ private void removeStaticModifierCodeActions(Diagnostic diagnostic, JavaCodeActi
...
+    /**
+     * Returns simple name for the given fully qualified name.
+     */
+    protected static String getSimpleName(String fqName) {
+        int idx = fqName.lastIndexOf('.');
+        if (idx != -1 && idx != fqName.length() - 1) {
+            return fqName.substring(idx + 1);
+        }
+        return fqName;
+    }
```
mrglavas, line 163: I recall seeing a method like this somewhere else in the project. Did you copy this from another class? We should be reusing code wherever possible instead of duplicating it.

mrglavas: Found this in `RemoveParamAnnotationQuickFix.java`. I did not check whether this is the only other copy of such a method.

```
    protected static String getShortName(String qualifiedName) {
        final int i = qualifiedName.lastIndexOf('.');
        if (i != -1) {
            return qualifiedName.substring(i+1);
        }
        return qualifiedName;
    }
```

anusreelakshmi934: Hi @mrglavas, I noticed that the `AbstractDiagnosticsCollector` class already has a `getSimpleName` method defined, so I updated the code to utilize `getSimpleName` from `AbstractDiagnosticsCollector` accordingly.

anusreelakshmi934: Also Do we need to update `RemoveParamAnnotationQuickFix.java` as well to use the `getSimpleName` method from `AbstractDiagnosticsCollector`? I also noticed that some classes are still using the pattern `annotationName.substring(annotationName.lastIndexOf('.') + 1)` to get the short name. Should we refactor those classes to consistently use `getSimpleName` for retrieving short names?

dessina-devasia (approved):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.java
mrglavas: @anusreelakshmi934 You should open a new issue describing your suggested improvements to reduce code duplication. This PR addresses issue #870 and I believe what you currently have here is sufficient for that.

mrglavas (approved):
Looks good to me, but @turkeylurkey should probably also review since these were his suggested improvements.

turkeylurkey (approved):
Thanks, I just wanted some other reviewer to validate my suggestion about the quotes etc.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/beanvalidation/BeanValidationQuickFix.java
anusreelakshmi934: Sure @mrglavas . I will open an Issue for the above suggestion to avoid code duplication. Thanks

PR #1091, opened 2024-11-11
Resolved inconsistent behaviour of @Dependency quickfix

### LINKED ISSUE #784: Jakarta EE: Inconsistent behaviour for @Dependent quick fix.
While reviewing the Jakarta EE quick fixes, I noticed a difference in behaviour in how the "Replace current scope with `@Dependent`" quick fix is applied when resolving a diagnostic on a field.

Current LT for IntelliJ behaviour:
Applying "Replace current scope with `@Dependent`" to a field adds the annotation to the field.
image
image

LT for VS Code 24.0.3 behaviour:
Applying "Replace current scope with `@Dependent`" to a field replaces the scope annotation on the class.
image
image

After applying the quick fix in VS Code it continues to report a diagnostic which cannot be resolved by the quick fix. We should investigate which of these tools (IntelliJ vs. VS Code) has the correct behaviour, assuming either of them are correct.

dessina-devasia:
Resolved inconsistent behaviour of @Dependency quickfix

Fixes #784

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanDiagnosticsCollector.java
@@ -80,7 +80,7 @@ public class ManagedBeanDiagnosticsCollector extends AbstractDiagnosticsCollecto
                 if (isManagedBean
                         && field.hasModifierProperty(PsiModifier.PUBLIC)
                         && !field.hasModifierProperty(PsiModifier.STATIC)
-                        && (fieldScopes.size() != 1 || !fieldScopes.get(0).equals(DEPENDENT_FQ_NAME))) {
+                        && !managedBeanAnnotations.contains(DEPENDENT_FQ_NAME)) {
                     diagnostics.add(createDiagnostic(field, unit,
                             Messages.getMessage("ManagedBeanWithNonStaticPublicField"),
                             DIAGNOSTIC_CODE, null,
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanQuickFix.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanQuickFix.java
@@ -17,8 +17,10 @@ package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.cdi;
 
 import com.intellij.openapi.progress.ProcessCanceledException;
 import com.intellij.openapi.project.IndexNotReadyException;
+import com.intellij.psi.PsiClass;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiModifierListOwner;
+import com.intellij.psi.util.PsiTreeUtil;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.JDTUtils;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.Messages;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.codeAction.proposal.quickfix.InsertAnnotationMissingQuickFix;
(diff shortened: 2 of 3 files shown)


vaisakhkannan (commented):
@dessina-devasia , Could you please share the screenshot after testing with this changes in the PR?

dessina-devasia:
> @dessina-devasia , Could you please share the screenshot after testing with this changes in the PR?

@vaisakhkannan please find the attached video having the latest behaviour

https://github.com/user-attachments/assets/c97152fe-a1ae-457c-ab8e-a48e491f1e90

vaisakhkannan (approved):
Thanks , Looks good to me based on the screen recording that dessina shared.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanDiagnosticsCollector.java
```diff
@@ -80,7 +80,7 @@ public void collectDiagnostics(PsiJavaFile unit, List<Diagnostic> diagnostics) {
                 if (isManagedBean
                         && field.hasModifierProperty(PsiModifier.PUBLIC)
                         && !field.hasModifierProperty(PsiModifier.STATIC)
-                        && (fieldScopes.size() != 1 || !fieldScopes.get(0).equals(DEPENDENT_FQ_NAME))) {
+                        && !managedBeanAnnotations.contains(DEPENDENT_FQ_NAME)) {
```
mrglavas, line 83: @dessina-devasia The change you have made here is not logically equivalent to the previous code. Is there a reason you removed the check on the size of the list? If there are multiple scope annotations that implies that one of them must be a scope other than `@Dependent` and from the definition above that would seem to be an error.

@turkeylurkey You've been spending a lot of time looking at this stuff lately. What do you think?

mrglavas (changes requested):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanDiagnosticsCollector.java
turkeylurkey: This code matches what we do in lsp4jakarta so it is best to keep this here. Michael will open an issue to describe an alternative way to handle this diagnostic and we can consider it for all the IDEs.

turkeylurkey (approved):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/cdi/ManagedBeanDiagnosticsCollector.java
dessina-devasia: @mrglavas The flag indication associated with the variable will persist without this fix. Additionally, I made these changes to align with the updates in Eclipse. The latest functionality will work as intended if these changes are included.

mrglavas (commented):
@dessina-devasia I spoke with @turkeylurkey and agreed that it's fine to move forward for now because this aligns exactly with LSP4Jakarta. I believe the user experience can be improved and will open an issue against LSP4Jakarta which describes that improvement.

In future, if you are copying and/or porting code from another project you should state that somewhere in the issue or PR. This helps the reviewer understand where the code came from and I would have reviewed this PR from the perspective of aligning the behaviour with LSP4Jakarta.

mrglavas (approved):
Approving, now that I understand the provenance of the code. Thanks.

dessina-devasia:
> Approving, now that I understand the provenance of the code. Thanks.

Thank you @mrglavas

dessina-devasia:
> @dessina-devasia I spoke with @turkeylurkey and agreed that it's fine to move forward for now because this aligns exactly with LSP4Jakarta. I believe the user experience can be improved and will open an issue against LSP4Jakarta which describes that improvement.
> 
> In future, if you are copying and/or porting code from another project you should state that somewhere in the issue or PR. This helps the reviewer understand where the code came from and I would have reviewed this PR from the perspective of aligning the behaviour with LSP4Jakarta.

Sure @mrglavas

PR #1093, opened 2024-11-12
Detect a SocketTimeoutException and exit the JUnit tests

### LINKED ISSUE #1121: Mac automated UI test failure: SocketTimeoutException, detect
The UI robot is trying to type characters into the editor but the editor process is not confirming the characters.
```
GradleSingleModJakartaLSTest > testJakartaQuickFixInJavaPart() FAILED
    java.net.SocketTimeoutException: Read timed out
        at java.base/sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:288)
        at java.base/sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:314)
        at java.base/sun.nio.ch.NioSocketImpl.read(NioSocketImpl.java:355)
        at java.base/sun.nio.ch.NioSocketImpl$1.read(NioSocketImpl.java:808)
        at java.base/java.net.Socket$SocketInputStream.read(Socket.java:966)
        at okio.InputStreamSource.read(JvmOkio.kt:94)
        at okio.AsyncTimeout$source$1.read(AsyncTimeout.kt:125)
        at okio.RealBufferedSource.indexOf(RealBufferedSource.kt:427)
        at okio.RealBufferedSource.readUtf8LineStrict(RealBufferedSource.kt:320)
        at okhttp3.internal.http1.Http1ExchangeCodec.readHeaderLine(Http1ExchangeCodec.java:242)
        ...
```
Detect that this exception has occurred and terminate testing in run.sh

turkeylurkey:
Fixes #1121 

When we examine the output logs of tests that experience a SocketTimeoutException we find that every subsequent test fails. Once the connection between the robot and the IDE is broken we cannot reconnect. 

The strategy is to save the JUnit output in a file and after each set of tests scan the file to see if the expected exception occurred. We use a static file reader so that it maintains its place in the file and we do not start scanning from the beginning each time. 

The benefit of detecting this exception is that we can exit the tests as soon as possible so the debugging can begin.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
@@ -562,4 +562,41 @@ public class TestUtils {
                 });
     }
 
+    /**
+     * Test to see if there has been a fatal error and JUnit should be stopped.
+     * This searches the output of this JUnit run for SocketTimeoutException which
+     * has been identified as a fatal error and occurs during the Mac tests.
+     * It is set up as a static reader so that we do not reread the whole file
+     * after each test.
+     */
+    final static String outputFile = System.getenv("JUNIT_OUTPUT_TXT");
+    final static BufferedReader reader;
+
+    static {
+        try {
+            reader = new BufferedReader(new InputStreamReader(new FileInputStream(outputFile)));
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void detectFatalError() {
+        try {
+            // Flush our output and then wait briefly for the process running 'tee' to flush output to the
+            // file that the buffered reader is reading.
+            System.out.flush();
+            Thread.sleep(10);
+            // Continue reading the existing reader
+            String line = null;
+            while ((line = reader.readLine()) != null) {
+                if (line.contains("SocketTimeoutException")) {
+                    System.exit(23);
+                }
+            }
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+        }
+    }
 }
--- a/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
@@ -1,3 +1,12 @@
+/*******************************************************************************
+ * Copyright (c) 2023, 2024 IBM Corporation.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
 package io.openliberty.tools.intellij.it;
 
 import org.junit.jupiter.api.Assertions;
--- a/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
@@ -582,14 +582,16 @@ public class TestUtils {
     final static BufferedReader reader;
 
     static {
+        BufferedReader reader1;
         try {
-            reader = new BufferedReader(new InputStreamReader(new FileInputStream(outputFile)));
+            reader1 = new BufferedReader(new InputStreamReader(new FileInputStream(outputFile)));
         } catch (FileNotFoundException e) {
-            throw new RuntimeException(e);
+            reader1 = null;
         }
+        reader = reader1;
     }
 
-    public static void detectFatalError() {
+    public static synchronized void detectFatalError() {
         try {
             // Flush our output and then wait briefly for the process running 'tee' to flush output to the
             // file that the buffered reader is reading.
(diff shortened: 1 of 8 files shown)


dessina-devasia (approved):
Looks good

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -562,4 +571,41 @@ public static void copyDirectory(String sourceDirectoryLocation, String destinat
...
+        try {
+            // Flush our output and then wait briefly for the process running 'tee' to flush output to the
+            // file that the buffered reader is reading.
+            System.out.flush();
+            Thread.sleep(10);
+            // Continue reading the existing reader
+            String line = null;
+            while ((line = reader.readLine()) != null) {
+                if (line.contains("SocketTimeoutException")) {
+                    System.exit(23);
```
mrglavas, line 602: This is a magic number: https://en.wikipedia.org/wiki/Magic_number_(programming). Does this exit code actually mean anything? Are the scripts looking for this value? I didn't see any references in this PR.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -562,4 +571,41 @@ public static void copyDirectory(String sourceDirectoryLocation, String destinat
...
+        }
+    }
+
+    public static void detectFatalError() {
+        try {
+            // Flush our output and then wait briefly for the process running 'tee' to flush output to the
+            // file that the buffered reader is reading.
+            System.out.flush();
+            Thread.sleep(10);
+            // Continue reading the existing reader
```
mrglavas, line 598: Does this actually work? Once you've reached the end of a stream, in general you can't read any more data from it. This is different from reading from a stream which blocks on read while it waits for the producer to generate more data.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -562,4 +571,41 @@ public static void copyDirectory(String sourceDirectoryLocation, String destinat
...
+
+    static {
+        try {
+            reader = new BufferedReader(new InputStreamReader(new FileInputStream(outputFile)));
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    public static void detectFatalError() {
```
mrglavas, line 592: synchronized? The reader has state and might break if more than one thread comes in here at a time.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: Would you like it better if I added a `public static int STE_FOUND = 23;` ?

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: I'm not sure how to test whether it works or not. Should I just take it out?

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -562,4 +571,41 @@ public static void copyDirectory(String sourceDirectoryLocation, String destinat
...
+     * after each test.
+     */
+    final static String outputFile = System.getenv("JUNIT_OUTPUT_TXT");
+    final static BufferedReader reader;
+
+    static {
+        try {
+            reader = new BufferedReader(new InputStreamReader(new FileInputStream(outputFile)));
+        } catch (FileNotFoundException e) {
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
turkeylurkey: Ok, I'll make it null and catch a null pointer exception later.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: Yes, that sounds good.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
mrglavas: What does taking it out mean? Replacing the "optimization" with a locally created Reader and avoiding this and the other issues I mentioned that come with managing it as a static field?

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: Actually, this value is not even passed by `gradle test` out to the environment. Gradle catches it and just exits with rc=1. So I think I'll just change this value to 1.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -562,4 +571,28 @@ public static void copyDirectory(String sourceDirectoryLocation, String destinat
                 });
     }
 
+    /**
+     * Test to see if there has been a fatal error and JUnit should be stopped.
+     * This searches the output of this JUnit run for SocketTimeoutException which
+     * has been identified as a fatal error and occurs during the Mac tests.
+     */
+    public static synchronized void detectFatalError() {
```
mrglavas, line 579: After moving the Reader's lifecycle into this method you don't need to synchronize this method. You may want to change this later, but wouldn't worry about doing it in this PR.

mrglavas (approved):
Looks good now. Thanks!

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
turkeylurkey: I will do this in a future update.

Issue #911, opened 2024-08-13
IntelliJ 2024.2: Migrate to IntelliJ Platform Gradle Plugin 2.x


TrevCraw:
As mentioned in Slack communication in the JetBrains Slack channel, plugins targeting 2024.2 must migrate from Gradle IntelliJ 1.x to IntelliJ Platform Gradle Plugin 2.x

More information on the migration can be found here: https://plugins.jetbrains.com/docs/intellij/tools-intellij-platform-gradle-plugin-migration.html

Link to Slack message: https://jetbrains-platform.slack.com/archives/C5U8BM1MK/p1721656635063899

staicy123:
Hi @TrevCraw,
Here I am attaching the commit with all the changes made to support 2024.2 version in my forked feature branch. I have completed FVT for 2024.2 with java version 17 gradle and once I complete the FVT for 2024.2 with java version 21 and 2024.1 java version 17. I will raise the PR for the same. https://github.com/OpenLiberty/liberty-tools-intellij/compare/main...staicy123:liberty-tools-intellij:%23911-Migrate-Gradle-Plugin-2.x-intellij-2024.2?expand=1
Thanks

staicy123:
Hi @TrevCraw, 
PR raised for the change here: https://github.com/OpenLiberty/liberty-tools-intellij/pull/1053.
Please review and share your comments.

Thanks

staicy123:
Hi @TrevCraw ,
I have updated the PR with all the changes required for supporting 2024.2 intellij version for running buildPlugin task.
https://github.com/OpenLiberty/liberty-tools-intellij/pull/1053
Unit test discussed here: https://github.com/OpenLiberty/liberty-tools-intellij/issues/1048 (Needs to be fixed from jetbrains end)
Ui test:
Two issues needs to be addressed for this one 

1. https://github.com/OpenLiberty/liberty-tools-intellij/issues/1088 (Started the analysis)
2. https://github.com/OpenLiberty/liberty-tools-intellij/issues/1018 (In progress fixing test failures)

The PR for  runIdeForUiTests task supporting intellij 2024.2 change in build.gradle will be raised by @vaisakhkannan once my changes are reviewed and merged.

Thanks

staicy123:
The changes have been successfully merged. Therefore, this ticket will be closed.

PR #1116, opened 2024-11-18
Removed unused java target version and added import for kotlinCompile

staicy123:
Part of : https://github.com/OpenLiberty/liberty-tools-intellij/issues/911
 The unused Java target version has been removed from gradle.properties, and the import for kotlinCompile has been added in build.gradle.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,6 @@ pluginSinceBuild=241
 pluginUntilBuild=242.*
 
 javaVersion=17
-javaTargetVersion=17
 
 # Target IntelliJ Community by default
 platformType=IC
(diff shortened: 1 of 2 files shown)


dessina-devasia:
Add Fixes/Part of corresponding to an issue in the description section

gradle.properties
```diff
@@ -3,7 +3,6 @@ pluginSinceBuild=241
 pluginUntilBuild=242.*
 
 javaVersion=17
-javaTargetVersion=17
```
dessina-devasia, line 6: Can you explain why javaTargetVersion removed?

staicy123: This property was initially added with the intention of specifying different source and target compatibility versions for Java. However, as both the source and target compatibility versions are now set to the same Java version, the property is no longer needed. Therefore, this line has been removed from gradle.properties.

dessina-devasia: Thanks for the clarification @staicy123

dessina-devasia (approved):
Looks good to me

vaisakhkannan (approved):
looks good , Thanks @staicy123

TrevCraw (approved):

Issue #1067, opened 2024-11-01
Intelij: 2024.2 Update Test Automation in Windows and Linux


vaisakhkannan:
Update UI test automation on Windows to support IntelliJ 2024.2.

- [x] SingleModJakartaLSTestCommon - [Done]
- [x] SingleModLibertyLSTestCommon - [Done]
- [x] SingleModMPLSTestCommon -[Done]
- [x] SingleModMPProjectTestCommon -[Done]
- [x] SingleModNLTRestProjectTestCommon - [Done]

vaisakhkannan:
**SingleModLibertyLSTestCommon** - [In Progress]
- [ ]  prepareEnv & cleanup
- [ ] testServerXMLFeatureHover
- [ ] testServerXMLNonFeatureHover
- [ ] testInsertFeatureIntoServerXML
- [ ] testInsertLibertyConfigElementIntoServerXML
- [ ] testInsertLibertyConfigIntoServerEnv
- [ ] testInsertLibertyConfigIntoBootstrapProps
- [ ] testServerEnvCfgHover
- [ ] testBootstrapPropsCfgHover
- [ ] testDiagnosticInServerXML
- [ ] testQuickFixInServerXML
- [ ] testDiagnosticInServerEnv
- [ ] testDiagnosticInBootstrapProperties

vaisakhkannan:
### SingleModJakartaLSTestCommon 

- [ ] testInsertJakartaCodeSnippetIntoJavaPart
- [ ] testJakartaDiagnosticsInJavaPart
- [ ] testJakartaQuickFixInJavaPart

vaisakhkannan:
### SingleModMPLSTestCommon

- [ ] testInsertCodeSnippetIntoJavaPart
- [ ] testMPDiagnosticsInJavaPart
- [ ] testMPQuickFixInJavaFile
- [ ] testInsertMicroProfileProperty
- [ ] testMicroProfileConfigHover
- [ ] testDiagnosticInMicroProfileConfigProperties
- [ ] testQuickFixInMicroProfileConfigProperties

vaisakhkannan:
### SingleModMPProjectTestCommon

- [ ] testOpenBuildFileActionUsingPopUpMenu
- [ ] testStartWithParamsActionUsingDropDownMenu
- [ ] testStartWithParamsActionUsingPlayToolbarButton
- [ ] testStartWithParamsActionUsingPopUpMenu
[...]

PR #1117, opened 2024-11-19
Issue #1088 - Parameterize the java target and platform target values

### LINKED ISSUE #1088: Configure UI tests to run with Java 21 and 2024.2
Determine how to parameterize the java target and platform target values and configure those values correctly for the default (production) build and for running the UI tests.
- default (production) build should target Java 17 and the latest 2024.1.* version
- the UI tests build should target Java 21 and the latest 2024.2.* version

anusreelakshmi934:
Fixes #1088 

By default, tasks like runIde use the IntelliJ Platform version specified in the dependencies extension. I implemented a solution by creating a new task called `runIdeLocally`, enabling us to set custom values for the Java and IntelliJ versions. This allows us to use `runIde` with the specified platform version (java 17 ) and Intellij version 2024.1.7 and `runIdeLocally` with the latest version (2024.2.2 in this case) and java 21.

Also updated the places where `runIde` is used in our code to `runIdeLocally`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -160,7 +160,7 @@ task copyDeps(type: Copy) {
 
 intellijPlatformTesting.runIde {
     runIdeLocally {
-        version = providers.gradleProperty("runIdeUiversion")
+        version = providers.gradleProperty("ideTargetVersion")
         type = IntelliJPlatformType.IntellijIdeaCommunity
         task {
             jvmArgumentProviders.add({
@@ -184,7 +184,7 @@ test {
 
 intellijPlatformTesting.runIde {
     runIdeForUiTests {
-        version = providers.gradleProperty("runIdeUiversion")
+        version = providers.gradleProperty("ideTargetVersion")
         type = IntelliJPlatformType.IntellijIdeaCommunity
         task {
             jvmArgumentProviders.add({
--- a/build.gradle
+++ b/build.gradle
@@ -177,7 +177,12 @@ test {
 
 intellijPlatformTesting.runIde {
     runIdeForUiTests {
+        version = providers.gradleProperty("runIdeUiversion")
         task {
+            // Java version for this task is set to Java 21
+            javaLauncher = javaToolchains.launcherFor {
+                languageVersion = JavaLanguageVersion.of(21)
+            }
             jvmArgumentProviders.add({
                 [
                         "-Drobot-server.port=8082",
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -157,13 +158,19 @@ task copyDeps(type: Copy) {
...
-runIde {
-    jvmArgumentProviders.add({
-        [
-                "--add-exports",
-                "java.base/jdk.internal.vm=ALL-UNNAMED",
-        ]
-    } as CommandLineArgumentProvider)
+intellijPlatformTesting.runIde {
+    runIdeLocally {
+        version = providers.gradleProperty("runIdeUiversion")
```
staicy123, line 163: Instead of using runIdeUiVersion, could you provide a more generic name that would be suitable for both the runIde and runIdeForUiTests tasks?

staicy123 (approved):
Approved, changes look good.

vaisakhkannan (commented):
@anusreelakshmi934 I can see in the Mac build that the IntelliJ Welcome frame is identified in the 23rd iteration. Do we need to increase the limit from 24?

Screenshot 2024-11-21 at 2 29 01 PM

anusreelakshmi934:
> @anusreelakshmi934 I can see in the Mac build that the IntelliJ Welcome frame is identified in the 23rd iteration. Do we need to increase the limit from 24?
> 
> Screenshot 2024-11-21 at 2 29 01 PM


This doesn’t seem to be a consistently occurring scenario. In some builds, it succeeds much earlier, even before the 10th iteration. For now, setting it to 24 should be fine. If we encounter issues in the future, we can consider increasing it further.

vaisakhkannan (approved):
looks good 


check my comment in the issue 1067 [here](https://github.com/OpenLiberty/liberty-tools-intellij/issues/1067#issuecomment-2491120510)

I ran the GHA build today, with the updated UI changes along with this PR changes. Now, I can see that both unit and UI tests are passing on Linux, while on Windows, all unit tests are passing, but 2 UI tests failed ( one is known issue and one is intermitted , if I rerun it will resolve the problem)

TrevCraw (commented):
Changes look good, but have one question. What exactly do you mean by this statement?

> we may need to update the runIde variable in our code to runIdeLocally, as we will be using runIdeLocally going forward.

vaisakhkannan (approved):

anusreelakshmi934:
> Changes look good, but have one question. What exactly do you mean by this statement?
> 
> > we may need to update the runIde variable in our code to runIdeLocally, as we will be using runIdeLocally going forward.

I meant renaming all instances where we use `runIde` to `runIdeLocally` . I have made that change

TrevCraw (approved):

Issue #1131, opened 2024-11-25
Intellij 2024.2: Extra Dialog Appears for Liberty Toolbar Actions for Multiple projects


vaisakhkannan:
Identified a behavioural difference in DevMode actions from the Liberty Tool window between IntelliJ versions `2024.1` and `2024.2` (the same behaviour is observed in 2024.3 as well).

An extra pop-up window appears to select the Liberty project, even though the dev mode of the corresponding project is already selected.

I found this behavior in the code of the feature branch `intellij-2024.2-support-backup`.

Behaviour in Intellij `2024.1`

https://github.com/user-attachments/assets/bf61cdd4-e3ab-432e-ab40-56c67f6a40ce

Behaviour in Intellij `2024.2`

https://github.com/user-attachments/assets/d33bb33b-d61e-4eac-9970-043175232c81

NOTE: Same project is tested in both intellij versions. 
The behaviour was observed today when I checked on Windows as part of the support Intellij 2024.3 FVT. Then, I checked on Mac with IntelliJ version 2024.2, and the same issue was present. During the initial FVT, we tested on Mac, where the project setup was specified as "Mac: generate projects to different folders." This is why we didn’t notice the issue at that time.

vaisakhkannan:
I observed a strange behavior: if we perform a right-click action on the project name and select the Liberty Dev Mode action, it executes properly. After this, everything works as expected. From the next time onward, the Liberty action performs correctly using the Liberty Play button without requiring the project to be selected, and no popup dialog appears.

To reproduce the issue again, uninstall the zip file and then reinstall it.

https://github.com/user-attachments/assets/2afeae90-2294-49c6-b966-c45106eb8ccb

anusreelakshmi934:
> I observed a strange behavior: if we perform a right-click action on the project name and select the Liberty Dev Mode action, it executes properly. After this, everything works as expected. From the next time onward, the Liberty action performs correctly using the Liberty Play button without requiring the project to be selected, and no popup dialog appears.
> 
> To reproduce the case mentioned in this issue again, uninstall the zip file and then reinstall it.
> 
>  Strange-behaviour.mov

I tried this scenario in linux too. It is reproducible

vaisakhkannan:
[...]

PR #1137, opened 2024-11-27
#1131 extra dialog appears for multiple projects

staicy123:
Fixes #https://github.com/OpenLiberty/liberty-tools-intellij/issues/1131
This PR addresses the issue where a behavioral inconsistency was identified in DevMode actions from the Liberty Tool window between IntelliJ versions 2024.1 and 2024.2 (with the same behavior observed in 2024.3). Specifically, an unnecessary pop-up window was appearing to select the Liberty project, even when the DevMode for the corresponding project was already selected.

The root cause of the issue was identified and resolved. The fix ensures that the extra pop-up no longer appears, streamlining the workflow for the affected IntelliJ versions.


src/main/java/io/openliberty/tools/intellij/util/TreeDataProvider.java
vaisakhkannan, line 1: @staicy123, please update header year from "2020, 2022" to "2020,  2024"

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunConfiguration.java
vaisakhkannan, line 1: @staicy123, please update header year from "2020, 2022" to "2020, 2024"

vaisakhkannan (changes requested):

src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunConfiguration.java
staicy123: Done

vaisakhkannan (dismissed):
Thanks @staicy123 , Changes looks good to me.

I have tested these changes using zip file generated in intellij versions 2024.1, 2024.2 and 2024.3. So works fine as expected in all 3 IDEs

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -55,7 +55,7 @@ public void actionPerformed(@NotNull AnActionEvent e) {
 
         // Obtain the liberty module associated to the current action.
         LibertyModule libertyModule = null;
-        VirtualFile buildFile = (VirtualFile) e.getDataContext().getData(Constants.LIBERTY_BUILD_FILE);
+        VirtualFile buildFile = e.getDataContext().getData(Constants.LIBERTY_BUILD_FILE);
```
turkeylurkey, line 58: Thank you for finding the problem but it is not clear how this code resolves it. Could you explain the problem that you found and how this fixes it? It appears that calling a different getData() is the solution, what does this do differently?

turkeylurkey (changes requested):

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
staicy123: In this comment: https://github.com/OpenLiberty/liberty-tools-intellij/issues/1131#issuecomment-2500084917, you can see an existing scenario where the build file is retrieved by right-clicking and selecting an action for a specific project without encountering a popup window for project selection.

Upon analyzing this behavior, I found that the getData method is invoked with a DataKey argument instead of a String. Further investigation revealed that the getData method with a String argument is marked as deprecated in 2024.3. While it is not marked as deprecated in 2024.2, it still does not work as expected. Typically, deprecated methods remain functional until fully removed, but in this case, the method is not behaving as intended.
Deprecated in 2024.3:
Screenshot 2024-11-28 at 9 57 11 AM

To resolve the issue, I adapted the working behavior observed in the right-click scenario (where actions are executed successfully) for other cases where the build file needs to be retrieved from the DataContext. This ensures we correctly capture the current state of the IDE when an action is invoked.
Screenshot 2024-11-28 at 9 58 42 AM

Let me know your thoughts!

Thanks

turkeylurkey: Ok, I agree with this analysis but the code is still hard to read. Usually a variable with a name like "LIBERTY_BUILD_FILE" is just a string or a file or even an integer. Could you rename the constant "LIBERTY_BUILD_FILE_DATAKEY" so it is easier to remember what it is when someone reads this line of code? Also as we discussed in the meeting a comment mentioning or summarizing this analysis.

staicy123: Thanks @turkeylurkey for the suggestions. I have updated the PR as per this review comment.

staicy123: As requested to see the usage of getData() method. I could see that the deprecated getData() method of DataContext Interface is called only once and I have added the comment there.
Thanks

vaisakhkannan (approved):

turkeylurkey (approved):
Looks good, thank you.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
turkeylurkey: 👍

PR #1148, opened 2024-12-03
Upgrade JDK to the latest version and use Apple Silicon architecture

### LINKED ISSUE #1132: Mac automated testing: use the correct version of IntelliJ
There is a message from IntelliJ that we are testing the x86 version of IntelliJ on Mac.
image

We are experiencing performance issues during testing on the Mac. We should make sure we are installing the native version of the software (on Mac that is Apple Silicon) to ensure we are getting the best performance we can get.

turkeylurkey:
Fixes #1132

IntelliJ started displaying a warning to use the Apple Silicon version of IntelliJ on Macs. When I tried that using JDK 17.0.6 the tests would hang. I upgraded to the latest JDK, 17.0.13, and the tests pass more often now than when we used the older JDK.


mrglavas (approved):
Looks good! Looking forward to more passing builds.

mrglavas (approved):

TrevCraw (approved):

PR #1199, opened 2024-12-18
Add constants for string literals in UI test classes

### LINKED ISSUE #1153: Add constants for string literals throughout the UI tests.
While reviewing https://github.com/OpenLiberty/liberty-tools-intellij/pull/1145 I noticed the UI tests have many string literals that are duplicated like "Compact Mode" and "Close All Tabs". These really should be moved to constants to make the code more maintainable.

anusreelakshmi934:
Fixes #1153

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/util/Constants.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/Constants.java
@@ -110,6 +110,5 @@ public final class Constants {
     public static final String SERVICE_LIVE_HEALTH_CHECK_JAVA = "ServiceLiveHealthCheck.java";
     public static final String MPG_PROPERTIES = "microprofile-config.properties";
 
-
-
+    public static final String CLOSE_ALL_TABS = "Close All Tabs";
 }
--- a/src/main/java/io/openliberty/tools/intellij/util/Constants.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/Constants.java
@@ -94,4 +94,18 @@ public final class Constants {
     public static final String JAKARTA_LANG_SERVER="Eclipse LSP4Jakarta";
     public static final String MICROPROFILE_SERVER="Eclipse LSP4MP";
 
+    /**
+     * Constants for Integration testing
+     */
+    public static final String SYSTEM_RESOURCE_JAVA = "SystemResource.java";
+    public static final String SYSTEM_RESOURCE_2_JAVA = "SystemResource2.java";
+    public static final String SYSTEM_RESOURCE = "SystemResource";
+    public static final String SYSTEM_RESOURCE_2 = "SystemResource2";
+    public static final String SERVER_XML = "server.xml";
+    public static final String SERVER_ENV = "server.env";
+    public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
+    public static final String COMPACT_MODE = "Compact Mode";
+
+
+
 }
--- a/src/main/java/io/openliberty/tools/intellij/util/Constants.java
+++ b/src/main/java/io/openliberty/tools/intellij/util/Constants.java
@@ -101,10 +101,14 @@ public final class Constants {
     public static final String SYSTEM_RESOURCE_2_JAVA = "SystemResource2.java";
     public static final String SYSTEM_RESOURCE = "SystemResource";
     public static final String SYSTEM_RESOURCE_2 = "SystemResource2";
+
     public static final String SERVER_XML = "server.xml";
     public static final String SERVER_ENV = "server.env";
     public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
     public static final String COMPACT_MODE = "Compact Mode";
+
+    public static final String SERVICE_LIVE_HEALTH_CHECK_JAVA = "ServiceLiveHealthCheck.java";
+    public static final String MPG_PROPERTIES = "microprofile-config.properties";
 
 
 
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
@@ -13,6 +13,7 @@ import com.automation.remarks.junit5.Video;
 import com.intellij.remoterobot.RemoteRobot;
 import com.intellij.remoterobot.fixtures.JTreeFixture;
 import io.openliberty.tools.intellij.it.fixtures.ProjectFrameFixture;
+import io.openliberty.tools.intellij.util.Constants;
 import org.junit.jupiter.api.*;
 
 import java.nio.file.Path;
@@ -61,8 +62,8 @@ public abstract class SingleModJakartaLSTestCommon {
     public static void cleanup() {
         ProjectFrameFixture projectFrame = remoteRobot.find(ProjectFrameFixture.class, Duration.ofMinutes(2));
 
-        UIBotTestUtils.closeFileEditorTab(remoteRobot, "SystemResource.java", "5");
-        UIBotTestUtils.closeFileEditorTab(remoteRobot, "SystemResource2.java", "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, Constants.SYSTEM_RESOURCE_JAVA, "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA, "5");
 
         UIBotTestUtils.closeProjectView(remoteRobot);
         UIBotTestUtils.closeProjectFrame(remoteRobot);
@@ -80,7 +81,7 @@ public abstract class SingleModJakartaLSTestCommon {
         String insertedCode = "public String methodname() {";
 
         // get focus on file tab prior to copy
-        UIBotTestUtils.clickOnFileTab(remoteRobot, "SystemResource.java");
+        UIBotTestUtils.clickOnFileTab(remoteRobot, Constants.SYSTEM_RESOURCE_JAVA);
 
         // Save the current content.
         UIBotTestUtils.copyWindowContent(remoteRobot);
@@ -90,8 +91,8 @@ public abstract class SingleModJakartaLSTestCommon {
 
         // Insert a code snippet into java part
         try {
-            UIBotTestUtils.insertCodeSnippetIntoSourceFile(remoteRobot, "SystemResource.java", snippetStr, snippetChooser);
-            Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", "SystemResource.java");
+            UIBotTestUtils.insertCodeSnippetIntoSourceFile(remoteRobot, Constants.SYSTEM_RESOURCE_JAVA, snippetStr, snippetChooser);
+            Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", Constants.SYSTEM_RESOURCE_JAVA);
             TestUtils.validateCodeInJavaSrc(pathToSrc.toString(), insertedCode);
         }
         finally {
@@ -109,28 +110,28 @@ public abstract class SingleModJakartaLSTestCommon {
         String privateString = "private Response getProperties() {";
         String flaggedString = "getProperties";
         String expectedHoverData = "Only public methods can be exposed as resource methods";
-        Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", "SystemResource2.java");
+        Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", Constants.SYSTEM_RESOURCE_2_JAVA);
 
         // get focus on file tab prior to copy
-        UIBotTestUtils.clickOnFileTab(remoteRobot, "SystemResource2.java");
+        UIBotTestUtils.clickOnFileTab(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA);
 
         // Modify the method signature
-        UIBotTestUtils.selectAndModifyTextInJavaPart(remoteRobot, "SystemResource2.java", publicString, privateString);
+        UIBotTestUtils.selectAndModifyTextInJavaPart(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA, publicString, privateString);
 
         try {
             // validate the method signature is no longer set to public
             TestUtils.validateStringNotInFile(pathToSrc.toString(), publicString);
 
             //there should be a diagnostic for "private" on method signature - move cursor to hover point
-            UIBotTestUtils.hoverInAppServerCfgFile(remoteRobot, flaggedString, "SystemResource2.java", UIBotTestUtils.PopupType.DIAGNOSTIC);
+            UIBotTestUtils.hoverInAppServerCfgFile(remoteRobot, flaggedString, Constants.SYSTEM_RESOURCE_2_JAVA, UIBotTestUtils.PopupType.DIAGNOSTIC);
 
             String foundHoverData = UIBotTestUtils.getHoverStringData(remoteRobot, UIBotTestUtils.PopupType.DIAGNOSTIC);
             TestUtils.validateHoverData(expectedHoverData, foundHoverData);
-            UIBotTestUtils.clickOnFileTab(remoteRobot, "SystemResource2.java");
+            UIBotTestUtils.clickOnFileTab(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA);
 
         } finally {
             // Replace modified content with the original content
-            UIBotTestUtils.selectAndModifyTextInJavaPart(remoteRobot, "SystemResource2.java", privateString, publicString);
+            UIBotTestUtils.selectAndModifyTextInJavaPart(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA, privateString, publicString);
         }
     }
 
@@ -144,24 +145,24 @@ public abstract class SingleModJakartaLSTestCommon {
         String privateString = "private Response getProperties() {";
         String flaggedString = "getProperties";
 
-        Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", "SystemResource2.java");
+        Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", Constants.SYSTEM_RESOURCE_2_JAVA);
         String quickfixChooserString = "Make method public";
 
         // get focus on file tab prior to copy
-        UIBotTestUtils.clickOnFileTab(remoteRobot, "SystemResource2.java");
+        UIBotTestUtils.clickOnFileTab(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA);
 
         // Save the current content.
         UIBotTestUtils.copyWindowContent(remoteRobot);
 
         // Modify the method signature
-        UIBotTestUtils.selectAndModifyTextInJavaPart(remoteRobot, "SystemResource2.java", publicString, privateString);
+        UIBotTestUtils.selectAndModifyTextInJavaPart(remoteRobot, Constants.SYSTEM_RESOURCE_2_JAVA, publicString, privateString);
 
         try {
             // validate public signature no longer found in java part
             TestUtils.validateStringNotInFile(pathToSrc.toString(), publicString);
 
             //there should be a diagnostic - move cursor to hover point
-            UIBotTestUtils.hoverForQuickFixInAppFile(remoteRobot, flaggedString, "SystemResource2.java", quickfixChooserString);
+            UIBotTestUtils.hoverForQuickFixInAppFile(remoteRobot, flaggedString, Constants.SYSTEM_RESOURCE_2_JAVA, quickfixChooserString);
 
             // trigger and use the quickfix popup attached to the diagnostic
             UIBotTestUtils.chooseQuickFix(remoteRobot, quickfixChooserString);
@@ -200,8 +201,8 @@ public abstract class SingleModJakartaLSTestCommon {
         // must be expanded here before trying to open specific files
         projTree.expand(projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
 
-        UIBotTestUtils.openFile(remoteRobot, projectName, "SystemResource", projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
-        UIBotTestUtils.openFile(remoteRobot, projectName, "SystemResource2", projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
+        UIBotTestUtils.openFile(remoteRobot, projectName, Constants.SYSTEM_RESOURCE, projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
+        UIBotTestUtils.openFile(remoteRobot, projectName, Constants.SYSTEM_RESOURCE_2, projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
 
 
         // Removes the build tool window if it is opened. This prevents text to be hidden by it.
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModLibertyLSTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModLibertyLSTestCommon.java
@@ -13,6 +13,7 @@ import com.automation.remarks.junit5.Video;
 import com.intellij.remoterobot.RemoteRobot;
 import com.intellij.remoterobot.fixtures.JTreeFixture;
 import io.openliberty.tools.intellij.it.fixtures.ProjectFrameFixture;
+import io.openliberty.tools.intellij.util.Constants;
 import org.junit.jupiter.api.*;
 
 import java.nio.file.Path;
@@ -59,11 +60,11 @@ public abstract class SingleModLibertyLSTestCommon {
      */
     @AfterAll
     public static void cleanup() {
-        UIBotTestUtils.closeFileEditorTab(remoteRobot, "server.xml", "5");
-        UIBotTestUtils.closeFileEditorTab(remoteRobot, "server.env", "5");
-        UIBotTestUtils.closeFileEditorTab(remoteRobot, "bootstrap.properties", "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, Constants.SERVER_XML, "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, Constants.SERVER_ENV, "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, Constants.BOOTSTRAP_PROPERTIES, "5");
         if (!remoteRobot.isMac()) {
-            UIBotTestUtils.runActionFromSearchEverywherePanel(remoteRobot, "Compact Mode", 3);
+            UIBotTestUtils.runActionFromSearchEverywherePanel(remoteRobot, Constants.COMPACT_MODE, 3);
         }
         UIBotTestUtils.closeProjectView(remoteRobot);
         UIBotTestUtils.closeProjectFrame(remoteRobot);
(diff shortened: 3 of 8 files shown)


src/main/java/io/openliberty/tools/intellij/util/Constants.java
```diff
@@ -94,4 +94,21 @@ public final class Constants {
...
+
+    public static final String SERVER_XML = "server.xml";
+    public static final String SERVER_ENV = "server.env";
+    public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
+    public static final String COMPACT_MODE = "Compact Mode";
+
+    public static final String SERVICE_LIVE_HEALTH_CHECK_JAVA = "ServiceLiveHealthCheck.java";
+    public static final String MPG_PROPERTIES = "microprofile-config.properties";
+
+    public static final String CLOSE_ALL_TABS = "Close All Tabs";
```
vaisakhkannan, line 113: @anusreelakshmi934 , based on Trevor suggestion, since constants are only used for the tests, then they should go in a file under the test directory.

vaisakhkannan (changes requested):

src/main/java/io/openliberty/tools/intellij/util/Constants.java
anusreelakshmi934: I have modified it. Please re review.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -541,7 +542,7 @@ public static boolean isServerStopNeeded(String wlpInstallPath) {
      */
     public static void checkDebugPort(String absoluteWLPPath, int debugPort) throws IOException {
         // Retrieve the WLP server.env file path
-        Path serverEnvPath = Paths.get(absoluteWLPPath, "wlp", "usr", "servers", "defaultServer", "server.env");
+        Path serverEnvPath = Paths.get(absoluteWLPPath, "wlp", "usr", "servers", "defaultServer", ItConstants.SERVER_ENV);
```
vaisakhkannan, line 545: @TrevCraw , Do we need to consider adding all those strings, such as 'wlp', 'usr', 'servers', etc., to the `ITConstants class`? if not then we can keep "server.env" here instead of `ItConstants.SERVER_ENV`

vaisakhkannan (changes requested):

dessina-devasia (commented):
Change Copyright year in all files

src/test/java/io/openliberty/tools/intellij/it/Utils/ItConstants.java
```diff
@@ -0,0 +1,22 @@
+package io.openliberty.tools.intellij.it.Utils;
```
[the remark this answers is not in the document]
anusreelakshmi934: Added the Copyrights header in ItConstants.java file

anusreelakshmi934:
> Change Copyright year in all files

Changed the Copyright year

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
TrevCraw: I don't think it would be beneficial to create a constant for each individual string. However, there is a lot of repetition when defining the various `Path` variables. It would be beneficial to create constants that contain a section of a path, for instance `wlp/usr/servers/defaultServer`, which could then be re-used elsewhere.

mrglavas (changes requested):
@anusreelakshmi934 I see several new failures that are occurring on the Linux and Windows builds, for example:

```
GradleSingleModJakartaLSTest > initializationError FAILED
    java.lang.RuntimeException: Unable to open file SystemResource
        at io.openliberty.tools.intellij.it.UIBotTestUtils.openFile(UIBotTestUtils.java:700)
        at io.openliberty.tools.intellij.it.SingleModJakartaLSTestCommon.prepareEnv(SingleModJakartaLSTestCommon.java:204)
        at io.openliberty.tools.intellij.it.GradleSingleModJakartaLSTest.setup(GradleSingleModJakartaLSTest.java:40)

        Caused by:
        java.util.NoSuchElementException: List is empty.
            at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:221)
            at com.intellij.remoterobot.fixtures.Fixture.findText(Fixture.kt:46)
            at io.openliberty.tools.intellij.it.UIBotTestUtils.openFile(UIBotTestUtils.java:687)
            ... 2 more
```
I suspect these might be related to recent changes in computing paths in this PR. Please verify.

dessina-devasia (approved):
Looks good to me. Thanks!

anusreelakshmi934:
> @anusreelakshmi934 I see several new failures that are occurring on the Linux and Windows builds, for example:
> 
> ```
> GradleSingleModJakartaLSTest > initializationError FAILED
>     java.lang.RuntimeException: Unable to open file SystemResource
>         at io.openliberty.tools.intellij.it.UIBotTestUtils.openFile(UIBotTestUtils.java:700)
>         at io.openliberty.tools.intellij.it.SingleModJakartaLSTestCommon.prepareEnv(SingleModJakartaLSTestCommon.java:204)
>         at io.openliberty.tools.intellij.it.GradleSingleModJakartaLSTest.setup(GradleSingleModJakartaLSTest.java:40)
> 
>         Caused by:
>         java.util.NoSuchElementException: List is empty.
>             at kotlin.collections.CollectionsKt___CollectionsKt.first(_Collections.kt:221)
>             at com.intellij.remoterobot.fixtures.Fixture.findText(Fixture.kt:46)
>             at io.openliberty.tools.intellij.it.UIBotTestUtils.openFile(UIBotTestUtils.java:687)
>             ... 2 more
> ```
> 
> I suspect these might be related to recent changes in computing paths in this PR. Please verify.

Hi @mrglavas . I have made the changes and now the test succeeds.

src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
```diff
@@ -198,10 +199,10 @@ public static void prepareEnv(String projectPath, String projectName) {
 
         // expand project directories that are specific to this test app being used by these testcases
         // must be expanded here before trying to open specific files
-        projTree.expand(projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
+        projTree.expand(projectName, ItConstants.SYSTEM_DIR_PATH);
 
-        UIBotTestUtils.openFile(remoteRobot, projectName, "SystemResource", projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
-        UIBotTestUtils.openFile(remoteRobot, projectName, "SystemResource2", projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
+        UIBotTestUtils.openFile(remoteRobot, projectName, ItConstants.SYSTEM_RESOURCE, projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
+        UIBotTestUtils.openFile(remoteRobot, projectName, ItConstants.SYSTEM_RESOURCE_2, projectName, "src", "main", "java", "io.openliberty.mp.sample", "system");
```
TrevCraw, line 205: Why not use `SYSTEM_DIR_PATH` for these two lines?

src/test/java/io/openliberty/tools/intellij/it/SingleModLibertyLSTestCommon.java
```diff
@@ -591,16 +592,16 @@ public static void prepareEnv(String projectPath, String projectName) {
         // get a JTreeFixture reference to the file project viewer entry
         ProjectFrameFixture projectFrame = remoteRobot.find(ProjectFrameFixture.class, Duration.ofMinutes(2));
         JTreeFixture projTree = projectFrame.getProjectViewJTree(projectName);
-        projTree.expand(projectName, "src", "main", "liberty", "config");
+        projTree.expand(projectName, ItConstants.CONFIG_DIR_PATH);
 
         // open server.xml file
-        UIBotTestUtils.openFile(remoteRobot, projectName, "server.xml", projectName, "src", "main", "liberty", "config");
+        UIBotTestUtils.openFile(remoteRobot, projectName, ItConstants.SERVER_XML, projectName, "src", "main", "liberty", "config");
```
TrevCraw, line 598: Why not use the `CONFIG_DIR_PATH` constant here (and below)?

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
```diff
@@ -344,7 +345,7 @@ public static void prepareEnv(String projectPath, String projectName) {
         JTreeFixture projTree = projectFrame.getProjectViewJTree(projectName);
 
         UIBotTestUtils.openFile(remoteRobot, projectName, "ServiceLiveHealthCheck", projectName, "src", "main", "java", "io.openliberty.mp.sample", "health");
-        UIBotTestUtils.openFile(remoteRobot, projectName, "microprofile-config.properties", projectName, "src", "main", "resources", "META-INF");
+        UIBotTestUtils.openFile(remoteRobot, projectName, ItConstants.MPG_PROPERTIES, projectName, "src", "main", "resources", "META-INF");
```
TrevCraw, line 348: Why not use `META_INF_DIR_PATH` here?

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -27,7 +28,7 @@ public class TestUtils {
     /**
      * WLP messages.log path.
      */
-    public static final Path WLP_MSGLOG_PATH = Paths.get("wlp", "usr", "servers", "defaultServer", "logs", "messages.log");
+    public static final Path WLP_MSGLOG_PATH = Paths.get(ItConstants.DEFAULT_SERVER_PATH, "logs", "messages.log");
```
TrevCraw, line 31: It would be good to create a constant for the `messages.log` path.

TrevCraw (changes requested):
It looks like a few spots to use the new constants may have been missed. I only marked the spots visible in the GitHub diffs. Please make sure to search through the test code to see if there are any other spots that were missed where the new constants can be used or where new constants should be created.

anusreelakshmi934:
> It looks like a few spots to use the new constants may have been missed. I only marked the spots visible in the GitHub diffs. Please make sure to search through the test code to see if there are any other spots that were missed where the new constants can be used or where new constants should be created.

Hi @TrevCraw . Initially, I replaced all occurrences where the new constants could be used. However, this caused a GitHub Actions (GHA) failure where the file could not be opened. Upon investigation, I discovered that the `openFile` method expects the `filePath` parameter in the format `String... filePath`. Replacing it with the constant resulted in an error because the input was not provided in the required `String...` format.

I also tried passing the value as a `String` array, but this led to a compilation issue. The problem arises because the method expects multiple strings as individual arguments and does not accept an array in place of these arguments.

Expected Output is  - 
image

But if we pass the constant we will get - 
image

If we pass the value as String array we will get - 
image

So the file will not be opened. Hence I kept the strings in openFile as it was.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
anusreelakshmi934: modified the code.

TrevCraw:
> If we pass the value as String array we will get - image
> 
> So the file will not be opened. Hence I kept the strings in openFile as it was.

Hi @anusreelakshmi934 , looking at the image above, it looks like you used `Arrays.toString()` for your `String` array parameter. That will not work, as `Arrays.toString()` returns a `String`, not an array of `Strings`. You should be able to pass an array of `Strings` to the `String...` varargs parameter.

If you take a step back and look at the original code, you will noticed the pattern for both `getPath()` and `openFile()` was always set up to work with `String...`. We should leave that format alone and adjust the newly created constants to match the existing code patterns. It would probably be best to modify your `String` constants that are paths with slashes to be `String` arrays instead.

anusreelakshmi934:
> > If we pass the value as String array we will get - image
> > So the file will not be opened. Hence I kept the strings in openFile as it was.
> 
> Hi @anusreelakshmi934 , looking at the image above, it looks like you used `Arrays.toString()` for your `String` array parameter. That will not work, as `Arrays.toString()` returns a `String`, not an array of `Strings`. You should be able to pass an array of `Strings` to the `String...` varargs parameter.
> 
> If you take a step back and look at the original code, you will noticed the pattern for both `getPath()` and `openFile()` was always set up to work with `String...`. We should leave that format alone and adjust the newly created constants to match the existing code patterns. It would probably be best to modify your `String` constants that are paths with slashes to be `String` arrays instead.

Thanks, @TrevCraw, for the clarification. Initially, when I passed the string array to the `openFile` method, it was being interpreted as the fifth parameter, with the project name as the fourth. This caused issues with passing the string array. To resolve this, I modified the approach by combining the `projectName` and the path. With this change, two of the classes are now passing, but others are still failing despite using the same approach. I am currently investigating the issue and will update the PR accordingly.

mrglavas:
> Thanks, @TrevCraw, for the clarification. Initially, when I passed the string array to the `openFile` method, it was being interpreted as the fifth parameter, with the project name as the fourth. This caused issues with passing the string array. To resolve this, I modified the approach by combining the `projectName` and the path. With this change, two of the classes are now passing, but others are still failing despite using the same approach. I am currently investigating the issue and will update the PR accordingly.

@anusreelakshmi934 I know you're currently working on updates to this PR so will wait for that before reviewing again. For the current state of the PR, I had similar feedback to what @TrevCraw already expressed. It should be possible to construct arrays from a combination of constants and variable values that are equivalent to the vararg syntax (for `String...`) that was in the original code.

mrglavas (changes requested):
See my recent comments. Awaiting changes before reviewing again.

dessina-devasia (dismissed):

src/test/java/io/openliberty/tools/intellij/it/Utils/ItConstants.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public static final String SYSTEM_RESOURCE_JAVA = "SystemResource.java";
+    public static final String SYSTEM_RESOURCE_2_JAVA = "SystemResource2.java";
+    public static final String SYSTEM_RESOURCE = "SystemResource";
+    public static final String SYSTEM_RESOURCE_2 = "SystemResource2";
+    public static final String SERVICE_LIVEHEALTH_CHECK = "ServiceLiveHealthCheck";
+
+    public static final String[] CONFIG_DIR_PATH = { "src", "main", "liberty", "config" };
+    public static final String[] SYSTEM_DIR_PATH = { "src", "main", "java", "io.openliberty.mp.sample", "system"};
+    public static final String[] HEALTH_DIR_PATH = { "src", "main", "java", "io.openliberty.mp.sample", "health" };
```
TrevCraw, line 25: Why use dots as part of the String values in the String arrays `SYSTEM_DIR_PATH` and `HEALTH_DIR_PATH`? The original hard coded values do not use any dots in most cases. Now you have to convert all the dots to slashes. Why not just remove the dots from the constants so you do not have to do that conversion each time?

Note, there are some spots - like https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java#L201 - where dots have to be used because of the way the UI robot expects the values. For these cases, we can use a different constant.

TrevCraw (changes requested):
There are lots of places where `String.join()` is used. Would it be possible to pass String arrays as was done before? You can build the String arrays from your existing constants dynamically if needed. Or, in most cases, you can create new constants that combine your existing constants.

For example, you have:
```
CONFIG_DIR_PATH = { "src", "main", "liberty", "config" };
SERVER_XML = "server.xml";
```

You can create a new constant that combines the two called `SERVER_XML_PATH` and then use that to pass as a parameter to the necessary methods.

anusreelakshmi934:
Hi @TrevCraw, 
I’ve addressed your comments and made the necessary changes. Please review them when you get a chance and let me know if this aligns with your expectations or if further modifications are needed. Thanks! 

Additionally, I’ve removed the [WLP_MSGLOG_PATH](https://github.com/OpenLiberty/liberty-tools-intellij/blob/034fa100e27273a30a07772d2aa550c01032b3e0/src/test/java/io/openliberty/tools/intellij/it/TestUtils.java#L30) from the TestUtils file and moved it to the ItConstants file as a slash-separated path. I felt that, instead of calling `Paths.get()`, it might be more efficient to directly specify the path there. Could you confirm if this approach is appropriate?

src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
```diff
@@ -90,8 +91,8 @@ public void testInsertJakartaCodeSnippetIntoJavaPart() {
 
         // Insert a code snippet into java part
         try {
-            UIBotTestUtils.insertCodeSnippetIntoSourceFile(remoteRobot, "SystemResource.java", snippetStr, snippetChooser);
-            Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "system", "SystemResource.java");
+            UIBotTestUtils.insertCodeSnippetIntoSourceFile(remoteRobot, ItConstants.SYSTEM_RESOURCE_JAVA, snippetStr, snippetChooser);
+            Path pathToSrc = Paths.get(projectsPath, projectName, ItConstants.SYSTEM_DIR_PATH, ItConstants.SYSTEM_RESOURCE_JAVA);
```
TrevCraw, line 95: Why is SYSTEM_DIR_PATH a String and not a String[]? It would be better to keep the format the same as the original code. You can always build a new String[] constant from String[] SYSTEM_DIR_PATH and String SYSTEM_RESOURCE_JAVA.

src/test/java/io/openliberty/tools/intellij/it/SingleModLibertyLSTestCommon.java
```diff
@@ -122,15 +123,15 @@ public void testInsertFeatureIntoServerXML() {
...
+        UIBotTestUtils.clickOnFileTab(remoteRobot, ItConstants.SERVER_XML);
 
         // Save the current server.xml content.
         UIBotTestUtils.copyWindowContent(remoteRobot);
 
         // Insert a new element in server.xml.
         try {
             UIBotTestUtils.insertStanzaInAppServerXML(remoteRobot, stanzaSnippet, 18, 40, UIBotTestUtils.InsertionType.FEATURE, true);
-            Path pathToServerXML = Paths.get(projectsPath, projectName, "src", "main", "liberty", "config", "server.xml");
+            Path pathToServerXML = Paths.get(projectsPath, projectName, ItConstants.CONFIG_DIR_PATH, ItConstants.SERVER_XML);
```
TrevCraw, line 134: Same comment here as for SYSTEM_DIR_PATH above. Would be better to keep as a String[].

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
```diff
@@ -95,8 +97,8 @@ public void testInsertCodeSnippetIntoJavaPart() {
 
         // Insert a code snippet into java part
         try {
-            UIBotTestUtils.insertCodeSnippetIntoSourceFile(remoteRobot, "ServiceLiveHealthCheck.java", snippetStr, snippetChooser);
-            Path pathToSrc = Paths.get(projectsPath, projectName, "src", "main", "java", "io", "openliberty", "mp", "sample", "health","ServiceLiveHealthCheck.java");
+            UIBotTestUtils.insertCodeSnippetIntoSourceFile(remoteRobot, ItConstants.SERVICE_LIVE_HEALTH_CHECK_JAVA, snippetStr, snippetChooser);
+            Path pathToSrc = Paths.get(projectsPath, projectName, ItConstants.HEALTH_DIR_PATH, ItConstants.SERVICE_LIVE_HEALTH_CHECK_JAVA);
```
TrevCraw, line 101: Same comment here as for SYSTEM_DIR_PATH and CONFIG_DIR_PATH above. Would be better to keep as a String[].

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
```diff
@@ -201,14 +203,14 @@ public void testInsertMicroProfileProperty() {
...
+        UIBotTestUtils.clickOnFileTab(remoteRobot, ItConstants.MPG_PROPERTIES);
 
         // Save the current microprofile-config.properties content.
         UIBotTestUtils.copyWindowContent(remoteRobot);
 
         try {
-            UIBotTestUtils.insertConfigIntoMPConfigPropertiesFile(remoteRobot, "microprofile-config.properties", cfgSnippet, cfgNameChooserSnippet, cfgValueSnippet, true);
-            Path pathToMpCfgProperties = Paths.get(projectsPath, projectName, "src", "main", "resources", "META-INF", "microprofile-config.properties");
+            UIBotTestUtils.insertConfigIntoMPConfigPropertiesFile(remoteRobot, ItConstants.MPG_PROPERTIES, cfgSnippet, cfgNameChooserSnippet, cfgValueSnippet, true);
+            Path pathToMpCfgProperties = Paths.get(projectsPath, projectName, String.join(File.separator, ItConstants.META_INF_DIR_PATH), ItConstants.MPG_PROPERTIES);
```
TrevCraw, line 213: Why do a String.join() here? The original code is passing a String[].

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
```diff
@@ -294,19 +296,19 @@ public void testQuickFixInMicroProfileConfigProperties() {
         String correctedValue = "mp.health.disable-default-procedures=true";
         String expectedHoverData = "Type mismatch: boolean expected. By default, this value will be interpreted as 'false'";
 
-        Path pathToMpCfgProperties = Paths.get(projectsPath, projectName,"src", "main", "resources", "META-INF", "microprofile-config.properties");
+        Path pathToMpCfgProperties = Paths.get(projectsPath, projectName, String.join(File.separator, ItConstants.META_INF_DIR_PATH), ItConstants.MPG_PROPERTIES);
```
TrevCraw, line 299: Same as above comment on String.join().

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -541,7 +540,7 @@ public static boolean isServerStopNeeded(String wlpInstallPath) {
      */
     public static void checkDebugPort(String absoluteWLPPath, int debugPort) throws IOException {
         // Retrieve the WLP server.env file path
-        Path serverEnvPath = Paths.get(absoluteWLPPath, "wlp", "usr", "servers", "defaultServer", "server.env");
+        Path serverEnvPath = Paths.get(absoluteWLPPath, ItConstants.DEFAULT_SERVER_PATH, ItConstants.SERVER_ENV);
```
TrevCraw, line 543: Same comment here as for SYSTEM_DIR_PATH, CONFIG_DIR_PATH and HEALTH_DIR_PATH above. Would be better to keep as a String[].

src/test/java/io/openliberty/tools/intellij/it/Utils/ItConstants.java
```diff
@@ -0,0 +1,43 @@
...
+    public static final String MESSAGES_LOG_PATH = "wlp/usr/servers/defaultServer/logs/messages.log";
+
+
+    public static final String SERVER_XML = "server.xml";
+    public static final String SERVER_ENV = "server.env";
+    public static final String BOOTSTRAP_PROPERTIES = "bootstrap.properties";
+    public static final String COMPACT_MODE = "Compact Mode";
+
+    public static final String SERVICE_LIVE_HEALTH_CHECK_JAVA = "ServiceLiveHealthCheck.java";
+    public static final String MPG_PROPERTIES = "microprofile-config.properties";
```
TrevCraw, line 40:
```suggestion
    public static final String MPCFG_PROPERTIES = "microprofile-config.properties";
```

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
```diff
@@ -60,7 +59,7 @@ public static void validateLibertyServerStopped(String testName, String wlpInsta
     public static void validateLibertyServerStopped(String testName, String wlpInstallPath, int maxAttempts, boolean failOnNoStop) {
         printTrace(TraceSevLevel.INFO, testName + ":validateLibertyServerStopped: Entry.");
 
-        String wlpMsgLogPath = Paths.get(wlpInstallPath, WLP_MSGLOG_PATH.toString()).toString();
+        String wlpMsgLogPath = Paths.get(wlpInstallPath, ItConstants.MESSAGES_LOG_PATH).toString();
```
TrevCraw, line 62: Since `wlpInstallPath` is a String and you have now changed `MESSAGES_LOG_PATH` to a String, there is no reason to use `Paths.get()` to convert the values to a Path and then convert back to a String with `toString()`. May as well just concatenate the Strings together to build the full path as a String.

TrevCraw (changes requested):
Hi @anusreelakshmi934 , please address the review comments below. And be sure to check the build results after making your changes.

It probably was not necessary to move `MESSAGES_LOG_PATH` to the `ItConstants` file since it is only used in one file. However, the change should be OK. Please take a look at one of the comments related to this change below.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
anusreelakshmi934: I have modified the code accordingly

src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
anusreelakshmi934: I updated the `SYSTEM_DIR_PATH` to a `String[]`. Additionally, I introduced new variables: `SERVER_XML_PATH`, `SERVER_ENV_PATH`, and `BOOTSTRAP_PROPERTIES_PATH`.

src/test/java/io/openliberty/tools/intellij/it/SingleModLibertyLSTestCommon.java
anusreelakshmi934: I have updated the SYSTEM_DIR_PATH to a String[]

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
anusreelakshmi934: I have updated the SYSTEM_DIR_PATH to a String[]

src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
anusreelakshmi934: I have removed the String.join() and combined the projectName and MPCFG path using combinePath method.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
anusreelakshmi934: I have updated it to String[].

src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
```diff
@@ -12,6 +12,7 @@
 import com.automation.remarks.junit5.Video;
 import com.intellij.remoterobot.RemoteRobot;
 import com.intellij.remoterobot.fixtures.JTreeFixture;
+import io.openliberty.tools.intellij.it.Utils.ItConstants;
```
turkeylurkey, line 15: I'd like to suggest using `import static io.openliberty.tools.intellij.it.Utils.ItConstants.*` so that you do not have to specify the class name every time you use a constant in all these test files. It gets repetitive.

src/test/java/io/openliberty/tools/intellij/it/TestUtils.java
TrevCraw: I do not see these changes in the latest diff

anusreelakshmi934: Hi @TrevCraw . Sorry for the confusion. I have made the MESSAGES_LOG_PATH a String[] similar to all the other constants. Thus I have kept it as Paths.get().

vaisakhkannan (approved):
Looks good

vaisakhkannan (changes requested):
@anusreelakshmi934 ,I think it's better to define constants for string literals present in the child test class as well.

vaisakhkannan (changes requested):

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModJakartaLSTest.java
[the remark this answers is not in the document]
anusreelakshmi934: updated the header

src/test/java/io/openliberty/tools/intellij/it/GradleSingleModMPProjectTest.java
```diff
@@ -43,8 +45,8 @@ public static void setup() {
         setSmMpProjResURI("api/resource");
         setSmMPProjOutput("Hello! Welcome to Open Liberty");
```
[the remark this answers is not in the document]
vaisakhkannan: Okay, since we agreed to address this in new follow-up issue, we can proceed with merging this PR.

vaisakhkannan (dismissed):
looks good

vaisakhkannan (dismissed):
approving again

vaisakhkannan (approved):

PR #1201, opened 2024-12-18
Add condition check to determine if the IntelliJ window is currently full screen

### LINKED ISSUE #1165: Windows Automated Test Failure: component must be showing on the screen
The automated tests maximize the IntelliJ window on a Windows runner because the screen is smaller than on Mac and Linux. The tests also use "minimize" to return the screen to normal size after the test is done. However if there is a problem then the minimize action may be applied when the IntelliJ window is not currently maximized (it is normal size). In this case the window is really minimized and is not displayed on the screen at all and we get the error below. 

```
INFO: 2024-12-03T22:06:19.160537300: Failed to process the Run tests action using Liberty tool window drop down (component must be showing on the screen to determine its location).
```

One solution is to keep track of the maximized state so we apply minimize properly. Another way is to use the maximize button on the upper right which toggles between maximize and normal. There may be other ways available in Windows O/S.

vaisakhkannan:
Fixes #1165

Updated the UI test to check whether the IntelliJ IDE is already in a `maximized` or `minimized` state before performing the `maximize` or `minimize` actions on the IntelliJ IDE window. 

This fix checks the size of the IntelliJ window against the screen size to decide if the window is maximized or not. If the window's width or height is greater than or equal to the screen size, it is considered full screen.

The reason for checking whether the window's width or height is greater than or equal to the screen size is that sometimes the IntelliJ window size exceeds the screen size. In such cases, it can be considered as full screen. It is not necessary for both the width and height of the IntelliJ IDE window to be greater than or equal to the screen size; if at least one dimension exceeds the screen size, the window can be considered as full screen.

If both the width and height of the IntelliJ IDE window are less than the screen size, then the window is not in full screen.

In the below screenshot:

During Maximization Method Call:

Dimension of the Windows OS screen: 1536x864
Dimension of the IntelliJ IDE window: 1550x830
In this case, the IntelliJ IDE width is greater than the screen size, so this is considered full screen.

During Minimization Method Call:

Dimension of the Windows OS screen: 1536x864
Dimension of the IntelliJ IDE window: 1400x776
In this case, both the width and height of the IntelliJ IDE window are less than the screen size, so the IDE is not in full screen.

I tested multiple scenarios by running the same and different tests without performing a cleanup. Everything worked as expected, even though maximizing the window was not performed. Minimizing the window worked correctly and vice versa.

Screenshot 2024-12-18 182613

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -13,6 +13,7 @@ import com.intellij.remoterobot.RemoteRobot;
 import com.intellij.remoterobot.fixtures.*;
 import com.intellij.remoterobot.fixtures.dataExtractor.RemoteText;
 import com.intellij.remoterobot.search.locators.Locator;
+import com.intellij.remoterobot.search.locators.Locators;
 import com.intellij.remoterobot.utils.Keyboard;
 import com.intellij.remoterobot.utils.RepeatUtilsKt;
 import com.intellij.remoterobot.utils.WaitForConditionTimeoutException;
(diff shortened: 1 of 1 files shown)


turkeylurkey (approved):

dessina-devasia (approved):
LGTM

PR #1156, opened 2024-12-05
#1127 mac UI test integration

staicy123:
Fixes [#1066](https://github.com/OpenLiberty/liberty-tools-intellij/issues/1066) and [#1127](https://github.com/OpenLiberty/liberty-tools-intellij/issues/1127)
UI Code Modifications for Mac:
* Updated the UI code to support IntelliJ versions 2024.3 and 2024.2.
* Introduced conditional logic to handle UI differences between these versions, ensuring compatibility with both.
* Added conditional checks to handle differences in the UI between Mac and Windows/Linux platforms.
Issue Resolutions During Integration:
* NullPointerException in detectFatalError Method:
    * While integrating changes from the main and intellij-2024.2-support-backup branches into the updated UI code, encountered a NullPointerException when running individual UI tests locally.
    * Root Cause: The detectFatalError method relies on the presence of the junit.out file in the build folder, which is generated during CI/CD pipeline execution via the ./gradlew test command (configured using export JUNIT_OUTPUT_TXT).
    * Resolution: Added a null check to handle cases where this file is not generated during local test execution.
* Heap OutOfMemoryError in GHA:
    * Encountered a Heap OutOfMemoryError while integrating the latest JDK changes in GitHub Actions (GHA) builds.
    * Resolution: Increased memory allocation to address the issue, ensuring successful builds in GHA.

PR #1208, opened 2025-01-07
Bring focus to Liberty toolbar before selecting Config

### LINKED ISSUE #1173: Automated Test Failure: testMultipleConfigEditHistory() Method Fails on some builds of the Cron Job Builds
I found that many of the builds have failed due to the test `testMultipleConfigEditHistory()`, which is the only test failing in most builds.

I am attaching the video of the test failure

anusreelakshmi934:
Fixes #1173 

I was not able to reproduce this locally. But From some of the videos, I observed that the configuration is already in a selected state. Clicking elsewhere and then returning to the configuration seems to resolve the issue. 
So I have added the changes to first click on the liberty toolbar and then click on the config toolbar.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -2188,6 +2188,10 @@ public class UIBotTestUtils {
         ProjectFrameFixture projectFrame = remoteRobot.find(ProjectFrameFixture.class, Duration.ofSeconds(10));
         ComponentFixture cfgSelectBox = projectFrame.getRunConfigurationsComboBoxButton();
 
+        // Click on the Liberty toolbar to give it focus.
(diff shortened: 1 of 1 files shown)


vaisakhkannan (commented):
@anusreelakshmi934 , I think we should wait for the PR build results and try running multiple OS instances at the same time to confirm if this fix resolves the mentioned issue. If possible, check the feasibility of adding a condition before executing the click on the Liberty toolbar, ensuring the condition is true only when the issue occurs.

vaisakhkannan:
@anusreelakshmi934 , Staicy attempted to address this issue in the merged PR https://github.com/OpenLiberty/liberty-tools-intellij/pull/1156/files. Try removing her update to the method `selectConfigUsingToolbar` in this PR and check the builds with only your changes.

anusreelakshmi934:
This change doesnt resolve the issue. Hence closing the PR

PR #1218, opened 2025-01-10
Found a possibility of limiting the Linux build only for the 24.0.9 LTI version

### LINKED ISSUE #1217: Check the possibility of limiting the Linux build only  for the 24.0.9 LTI version.
Since we are getting a green build on `Linux OS` for the 24.0.9 LTI version, we should check the possibility of limiting the Linux build to run GHA builds only against the `24.0.9` LTI version.

vaisakhkannan:
Fixes #1217

This fix added a check (based on the value 'excludeFromBuild') on macOS and Windows to verify if 'refLTITag' is equal to '24.0.9'. The value of 'excludeFromBuild' is used to determine whether the steps below (Steps in Job `build`) need to be executed. This 'excludeFromBuild' variable can be removed once the '24.0.9' version is removed from the 'tag' array of the cron job build.

 I checked other approaches and I found that, there is no direct way to limit the OS (Linux only) for running the cron job builds specifically to a particular LTI version. We are just passing the tag array from cron job builds and calling build.yaml from cronjob.yaml file. So we can't change anything in cronjob.yaml file. So I tried this solution to make the changes in build.yaml file.
 
 Below are the Github documentations that I referred  as part of investigation.
 
 https://docs.github.com/en/enterprise-cloud@latest/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixexclude
 
 https://docs.github.com/en/enterprise-cloud@latest/actions/sharing-automations/reusing-workflows#using-a-matrix-strategy-with-a-reusable-workflow
 
 
I checked other approaches as well to avoid adding an if condition for all the steps, but they didn't work. I tried adding a single step with code to skip all the remaining steps, but that didn't work either. I also checked GitHub documentation but didn't find anything related to this.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,7 +55,8 @@ jobs:
           - runtime: linux
             os: ubuntu-latest
             reportName: linux-test-report
-            # Adding a check on macOS and Windows to identify refLTITag is equal to 24.0.9 , The value of 'excludeFromBuild' is use to run the steps
+            # Adding a check on macOS and Windows to verify if 'refLTITag' is equal to '24.0.9'. The value of 'excludeFromBuild' is used to determine whether the steps below need to be executed.
+            # This 'excludeFromBuild' variable can be removed once the '24.0.9' version is removed from the 'tag' array of the cron job build.
           - runtime: mac
             os: macOS-latest
             reportName: mac-test-report
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -55,12 +55,15 @@ jobs:
           - runtime: linux
             os: ubuntu-latest
             reportName: linux-test-report
+            # Adding a check on macOS and Windows to identify refLTITag is equal to 24.0.9 , The value of 'excludeFromBuild' is use to run the steps
           - runtime: mac
             os: macOS-latest
             reportName: mac-test-report
+            excludeFromBuild: ${{ inputs.refLTITag == '24.0.9' }}
           - runtime: windows
             os: windows-latest
             reportName: windows-test-report
+            excludeFromBuild: ${{ inputs.refLTITag == '24.0.9' }}
     env:
       USE_LOCAL_PLUGIN: ${{ inputs.useLocalPlugin || false }}
       REF_LSP4IJ: ${{ inputs.refLsp4ij }}
@@ -68,44 +71,47 @@ jobs:
       REF_LTI_TAG: ${{ inputs.refLTITag }}
     steps:
       - name: Configure pagefile
-        if: contains(matrix.os, 'windows')
+        if: ${{ contains(matrix.os, 'windows') && !matrix.excludeFromBuild }}
         uses: al-cheb/configure-pagefile-action@v1.4
         with:
           minimum-size: 8GB
           maximum-size: 10GB
           disk-root: "C:"
       - name: 'Checkout liberty-tools-intellij'
+        if: ${{ !matrix.excludeFromBuild }}
         uses: actions/checkout@v4
         with:
           path: liberty-tools-intellij
           ref: ${{ env.REF_LTI_TAG }}
       - name: 'Install required integration test software'
+        if: ${{ !matrix.excludeFromBuild }}
         working-directory: ./liberty-tools-intellij
         run: bash ./src/test/resources/ci/scripts/setup.sh
 
       # Checkout and build lsp4ij only if USE_LOCAL_PLUGIN is true
       - name: 'Checkout lsp4ij'
-        if: ${{ inputs.useLocalPlugin == true }}
+        if: ${{ inputs.useLocalPlugin == true && !matrix.excludeFromBuild }}
         uses: actions/checkout@v4
         with:
           repository: redhat-developer/lsp4ij
           path: lsp4ij
           ref: ${{ env.REF_LSP4IJ }}
       - name: 'Build Lsp4ij'
-        if: ${{ inputs.useLocalPlugin == true }}
+        if: ${{ inputs.useLocalPlugin == true && !matrix.excludeFromBuild }}
         working-directory: ./lsp4ij
         run: bash ./gradlew buildPlugin
       - name: 'Unzip lsp4ij file'
-        if: ${{ inputs.useLocalPlugin == true }}
+        if: ${{ inputs.useLocalPlugin == true && !matrix.excludeFromBuild }}
         working-directory: ./lsp4ij/build/distributions
         run: |
           unzip -o '*.zip' -d .
 
       - name: 'Build Liberty-Tools-Intellij'
+        if: ${{ !matrix.excludeFromBuild }}
         working-directory: ./liberty-tools-intellij
         run: bash ./gradlew buildPlugin -PuseLocal=${{ env.USE_LOCAL_PLUGIN }}
       - name: 'Archive artifacts'
-        if: ${{ runner.os == 'Linux' && !failure() }}
+        if: ${{ (runner.os == 'Linux' && !failure()) && !matrix.excludeFromBuild }}
         uses: actions/upload-artifact@v4.3.4
         with:
           name: liberty-tools-intellij-LTI-${{ env.REF_LTI_TAG || 'default' }}-LSP4IJ-${{ env.LSP4IJ_BRANCH }}
@@ -115,11 +121,12 @@ jobs:
           if-no-files-found: warn
           retention-days: 7
       - name: 'Run UI integration tests'
+        if: ${{ !matrix.excludeFromBuild }}
         id: run_tests
         working-directory: ./liberty-tools-intellij
         run: bash ./src/test/resources/ci/scripts/run.sh
       - name: 'Archive Test logs and reports'
-        if: ${{ failure() && steps.run_tests.conclusion == 'failure' }}
+        if: ${{ (failure() && steps.run_tests.conclusion == 'failure') && !matrix.excludeFromBuild }}
         uses: actions/upload-artifact@v4.3.4
         with:
           name: ${{ matrix.reportName }}-LTI-${{ env.REF_LTI_TAG || 'default' }}-LSP4IJ-${{ env.LSP4IJ_BRANCH }}


vaisakhkannan:
Build Resullts example : https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12704625074

https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12704635529

NOTE: You can see that for the LTI tag 24.0.9, both Windows and Mac have green builds because all the steps were skipped.

anusreelakshmi934 (approved):
Can we include only the Linux matrix and exclude the other two? Is it possible to modify the code so that the condition doesn't need to be checked in every step?

vaisakhkannan:
Hi @anusreelakshmi934 and @TrevCraw ,
I have updated the PR based on your suggestions and below are the build results , please check

https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12742932719 -> Normal Build
https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12743353818

https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12742954503 -> Cron Job builds
https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12743638005

NOTE: Only linux os is listed in the 24.0.9 LTI tag builds and you can see the change in name due to matrix constraints, please let me know the suggestions if any,

.github/workflows/build.yaml
```diff
@@ -50,17 +50,17 @@ jobs:
...
-        runtime: [ linux, mac, windows ]
-        include:
-          - runtime: linux
-            os: ubuntu-latest
-            reportName: linux-test-report
-          - runtime: mac
-            os: macOS-latest
-            reportName: mac-test-report
-          - runtime: windows
+        os: [ ubuntu-latest, macOS-latest, windows-latest ]
```
TrevCraw, line 53: Is there a reason the matrix needed to be changed up? Could we not just exclude `runtime: windows` or `runtime: mac`?

.github/workflows/build.yaml
```diff
@@ -50,17 +50,17 @@ jobs:
...
+        # This code can be removed once the '24.0.9' version is removed from the 'tag' array in the cron job build, and the previous code can be added back.
+        is24.0.9:
+          - ${{ inputs.refLTITag == '24.0.9' }}
+        # Adding a check on macOS and Windows to verify if 'is24.0.9' is equal to 'true'. The value of 'is24.0.9' is used to determine whether the job need to be executed.
+        exclude:
+          - is24.0.9: true
             os: windows-latest
-            reportName: windows-test-report
+          - is24.0.9: true
+            os: macOS-latest
```
mrglavas, line 63: Wondering if this can be made more generic to give the build more flexibility on what we're running without having to change the code each time.

mrglavas (commented):
See review comments.

.github/workflows/build.yaml
turkeylurkey, line 53: In the past if we wanted to run Mac twice we could just add "mac1, mac2" to the matrix.runtime. With this change it doesn't look like we can do that anymore. Can you provide instructions how to run a test multiple times or put the matrix back the way it was?

turkeylurkey (changes requested):

.github/workflows/build.yaml
vaisakhkannan: Yes @turkeylurkey , you can add multiple OS options to the OS matrix as shown below.

`os: [ ubuntu-latest, macOS-latest, windows-latest, ubuntu-latest, macOS-latest, windows-latest ]`

Normal build Screenshot

Screenshot 2025-01-14 at 10 05 36 AM

vaisakhkannan: Yes @TrevCraw ,I tried keeping the `runtime`, but I am getting the error below because I am using `exclude` to limit the OS for `24.0.9`.

Screenshot 2025-01-14 at 11 43 15 AM

https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12761300513

.github/workflows/build.yaml
vaisakhkannan: @mrglavas , 

I renamed the variable `is24.0.9` to `excludeLTITag`. Is that what you were expecting? 

There is no need to update the code each time. That means, If we remove `24.0.9` from the LTI tags in the `cronJob.yaml` file, there is no need to modify anything in the` build.yaml` file. The variable `excludeLTITag` will calculate its value  and will return `false` always since it not there in LTI tag. This ensures it works as expected for both normal and cron job builds.

In the future, if there is a need to limit the LTI tag `24.0.12` to Linux only, the only change required in the build.yaml file is to update `24.0.9` to `24.0.12`.

For cases where the LTI tag `24.0.12` should be limited to `macOS` only, the exclude section can be updated as follows:

```
exclude:
  - excludeLTITag: true
    os: ubuntu-latest
  - excludeLTITag: true
    os: windows-latest

```

NOTE: Whatever we are currently running for normal builds will continue to work as expected, even after removing the specified `excludeLTITag` from the LTI tag array in the cron job build.

It is not necessary to specify the `include` which is currently present in the [main](https://github.com/OpenLiberty/liberty-tools-intellij/blob/75d3bf77176a90acd45ccdf8b5069fe299073cd3/.github/workflows/build.yaml#L54C9-L63C44) code because it will be taken by default based on the OS specified in the matrix. In this case, exclude performs the filtering, while other cases work fine.

.github/workflows/build.yaml
turkeylurkey: Will the test report names collide?

vaisakhkannan: @TrevCraw , I tried a different approach today that is working fine by keeping the `runtime: mac`. This allows adding multiple OS options anytime by giving them different names and assigning unique names to the artifacts (both zip and test-reports)

vaisakhkannan: Yes you are correct @turkeylurkey , I tried a different approach today that is working fine. This allows adding multiple OS options anytime by giving them different names and assigning unique names to the artifacts (both zip and test-reports)

For example

```
jobs:
  build:
    name: build ( ${{ matrix.runtime.name }}, ${{ matrix.excludeLTITag }} )
    runs-on: ${{ matrix.runtime.os }}
    strategy:
      fail-fast: false
      matrix:
        runtime:
          - name: linux-1
            os: ubuntu-latest
            reportName: linux-test-report-1
          - name: mac-1
            os: macOS-latest
            reportName: mac-test-report-1
          - name: windows-1
            os: windows-latest
            reportName: windows-test-report-1
          - name: linux-2
            os: ubuntu-latest
            reportName: linux-test-report-2
          - name: mac-2
            os: macOS-latest
            reportName: mac-test-report-2
          - name: windows-2
            os: windows-latest
            reportName: windows-test-report-2
        excludeLTITag:
          - ${{ inputs.refLTITag == '24.0.9' }}
        exclude:
          - excludeLTITag: true
            runtime:
              name: mac-1
          - excludeLTITag: true
            runtime:
              name: mac-2
          - excludeLTITag: true
            runtime:
              name: windows-1
          - excludeLTITag: true
            runtime:
              name: windows-2
```

vaisakhkannan:
Latest build results: https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/12782956248 -Cron Job build

.github/workflows/build.yaml
mrglavas: @vaisakhkannan This is better though I was thinking about the possibility of parameterizing the values (e.g. 24.0.9) so that they are an input to the script rather than being embedded within. I understand from @TrevCraw that there are limitations to GitHub actions, specifically what is allowed in declaring a matrix that may not allow for such a solution.

vaisakhkannan: @mrglavas ,Do you mean adding `24.0.9` as a `workflow_call` input? In that case, we would need to add the value `24.0.9` twice in the `cronjob.yam`l file, and we would also need to pass those values during workflow calls, which could lead to code duplication. Adding `24.0.9` as a `workflow_dispatch` input won't trigger the daily Cron Job build. So, I believe that's not necessary.

mrglavas: @vaisakhkannan I wasn't sure of the details but I guess that would have meant adding another input to the `workflow_call` as you've described. Thanks for clarifying.

vaisakhkannan: @mrglavas , Just to confirm, So there’s no need to add an extra workflow call input for 24.0.9, right?

mrglavas: > @mrglavas , Just to confirm, So there’s no need to add an extra workflow call input for 24.0.9, right?

@vaisakhkannan Yes, that's right.

TrevCraw (dismissed):

TrevCraw (commented):
The changes seem OK to me, but I'm confused by the shared results. What is linux-1, linux-2, etc? Why are there multiple builds of the same OS? Is that just testing something out or are the builds now like that by default?

vaisakhkannan:
> Is that just testing something out or are the builds now like that by default?

@TrevCraw , No, that's not the default. I added linux-1 and linux-2 for the same OS to verify that the builds are executing properly and to ensure that artifact names are unique for each runtime, preventing any name collisions in the reports. We also have scenarios that require testing by running multiple runtime instances simultaneously for mutiple os.I have checked that possibility and its working. Currently, in this PR, there is only one runtime instance per OS

docs/LSP4IJ-Continuous-Integration.md
```diff
@@ -306,4 +309,25 @@ By following these steps, this job ensures that a notification is sent to Slack
...
\ No newline at end of file
+![Build Results in Slack](images/build-results-slack-view.png)
+
+# How to specify an LTI tag to limit OS in CI-CD Cron Job builds
+
+In order to limit the **OS** to run the builds against the **LTI** Tag, the **LTI** Tag must be specified in one location within the **build.yaml** file. This is used in the `build` job. There is a variable `excludeLTITag` under the `matrix` that is used to check the **LTI** Tag and confirm that the exact one is coming from the Cron Job call.
+
+![specify LTI tag to limit os](images/specify-LTI-tag-to-limit-os.png)
+
+There is no need to update the code each time. That means, If we remove `24.0.9` from the **LTI** tags in the **cronJob.yaml** file, there is no need to modify anything in the **build.yaml** file. The variable `excludeLTITag` will calculate its value and will return `false` always since it not there in **LTI** tag. This ensures it works as expected for both **normal** and **cron job** builds.
```
turkeylurkey, line 320: since it **is** not there

.github/workflows/build.yaml
turkeylurkey: Ok, I think people will understand the latest code that uses an array of objects under the `runtime` object.

turkeylurkey:
> What is linux-1, linux-2, etc? 
@TrevCraw I think that was answering my question about running ten linux builds when you want to reproduce an intermittent error.

docs/LSP4IJ-Continuous-Integration.md
vaisakhkannan: I updated the change you suggested, Thanks for the suggestion

anusreelakshmi934:
@vaisakhkannan could you please add the link to the latest build here

vaisakhkannan:
@anusreelakshmi934 , I resolved the merge conflicts of this PR today. 

Latest Build Results: 
https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/13026221632  -- Cron Job build
https://github.com/OpenLiberty/liberty-tools-intellij/actions/runs/13027868635?pr=1218 - > Normal Build

https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/13026250897 -- LSP4IJ PR number 777 & LTI Main

.github/workflows/build.yaml
```diff
@@ -168,6 +181,6 @@ jobs:
         if: ${{ failure() && steps.run_tests.conclusion == 'failure' }}
         uses: actions/upload-artifact@v4.3.4
         with:
-          name: ${{ matrix.reportName }}-LTI-${{ env.REF_LTI_TAG || 'default' }}-LSP4IJ-${{ env.LSP4IJ_BRANCH || 'default' }}
+          name: ${{ matrix.runtime.reportName }}-LTI-${{ env.REF_LTI_TAG || 'default' }}-LSP4IJ-${{ env.LSP4IJ_BRANCH || 'default' }}
```
dessina-devasia, line 184: @vaisakhkannan Can you explain the reason of adding runtime here?

.github/workflows/build.yaml
```diff
@@ -168,6 +181,6 @@ jobs:
...
           path: |
-            liberty-tools-intellij/build/reports/
+            liberty-tools-intellij/build/reports/
```
dessina-devasia, line 186: No change in this section, seems an extra line added here

.github/workflows/build.yaml
vaisakhkannan: @dessina-devasia , There is a change in the matrix strategy with this PR.
Previously, the code was:

```
matrix:
  runtime: [linux, mac, windows]
  include:
    - runtime: linux
      os: ubuntu-latest
      reportName: linux-test-report
    - runtime: mac
      os: macOS-latest
      reportName: mac-test-report
    - runtime: windows
      os: windows-latest
      reportName: windows-test-report
```

With this PR, the goal is to retrieve the unique `reportName` using only via `matrix.runtime.reportName`.

.github/workflows/build.yaml
vaisakhkannan: It seems like there is a bug from GitHub, I could not see an extra line.

turkeylurkey: Most likely you added a carriage return at the end of the line. Or removed one. It is highly unlikely there is a bug with github.

Issue #1202, opened 2024-12-19
java.lang.NullPointerException occurs when adding a configuration to a newly created project.


anusreelakshmi934:
![image](https://github.com/user-attachments/assets/a8ea0c31-76b0-4b5f-87ef-e379d771ce58)

Steps to Reproduce:
1. Run the LTI extension
2. Open a new project
3. Create a new Liberty run/debug config through the “Run” → “Edit configurations” window in IntelliJ
4. Add any param and then try to run the configuration.

The dev mode doesnt get started and we will encounter the exception.
However, once we start using the default Start action without any configuration, stop it, and then try starting again with the configuration, it works.

Verified the same issue occurs with 24.0.9 version also.

<details><summary>Log</summary>
<p>
java.lang.NullPointerException: Cannot invoke "com.intellij.execution.impl.SingleConfigurationConfigurable.getConfiguration()" because "selected" is null
	at com.intellij.execution.impl.EditConfigurationsDialog$2.actionPerformed(EditConfigurationsDialog.java:137)
	at java.desktop/javax.swing.AbstractButton.fireActionPerformed(AbstractButton.java:1972)
	at java.desktop/javax.swing.AbstractButton$Handler.actionPerformed(AbstractButton.java:2314)
	at java.desktop/javax.swing.DefaultButtonModel.fireActionPerformed(DefaultButtonModel.java:407)
	at java.desktop/javax.swing.DefaultButtonModel.setPressed(DefaultButtonModel.java:262)
	at java.desktop/javax.swing.plaf.basic.BasicButtonListener.mouseReleased(BasicButtonListener.java:279)
	at java.desktop/java.awt.AWTEventMulticaster.mouseReleased(AWTEventMulticaster.java:297)
	at java.desktop/java.awt.Component.processMouseEvent(Component.java:6662)
	at java.desktop/javax.swing.JComponent.processMouseEvent(JComponent.java:3394)
	at java.desktop/java.awt.Component.processEvent(Component.java:6427)
	at java.desktop/java.awt.Container.processEvent(Container.java:2266)
	at java.desktop/java.awt.Component.dispatchEventImpl(Component.java:5032)
	at java.desktop/java.awt.Container.dispatchEventImpl(Container.java:2324)
	at java.desktop/java.awt.Component.dispatchEvent(Component.java:4860)
	at java.desktop/java.awt.LightweightDispatcher.retargetMouseEvent(Container.java:4963)
[...]

PR #1235, opened 2025-01-17
Extra rework for issue 1202

### LINKED ISSUE #1234: Synchronize access to Liberty Modules singleton
The `LibertyModules.addLibertyModule()` method updates the local map variable in a non-atomic way so we should add a `synchronize` keyword like we see in some of the other methods in that class.

turkeylurkey:
Fixes #1234 

Three code improvements were suggested in the review of issue #1202.

1. Return `this` instead of calling `getInstance()`
2. Synchronize access to `libertyModules` in `addLibertyModule()`
3. Use an enum instead of Strings for project type in the area of Liberty actions. Using an enum allows us to remove some string checking code that is no longer required. This affects a lot of files but it makes the code more robust.


mrglavas (approved):
Looks good. Thanks!

vaisakhkannan (approved):
Looks good , Thanks @turkeylurkey

PR #1247, opened 2025-01-29
Fixed the retrieval of the local LSP4IJ path.

### LINKED ISSUE #1242: The LSP4IJ version is not being retrieved correctly when specifying the local path.
In our build.gradle file we have defined lsp4ij version as follows :

```
def lsp4ij = providers.gradleProperty('useLocal') && providers.gradleProperty('useLocal') == 'true' ?
                file("../lsp4ij/build/distributions/LSP4IJ/") :
                'com.redhat.devtools.lsp4ij:' + lsp4ijVersion
        plugins lsp4ij
```
As per the latest changes in the `build.gradle` file during the migration to version 2.x, the LSP4IJ version isn't being fetched correctly. Despite running with `-PuseLocal=true`, it's still using the version specified in the `build.gradle` file.

I verified this by running the extension from an older LTI branch that uses the Gradle 1.x version, where the locally built LSP4IJ plugin is resolved correctly when `-PuseLocal=true` is set. However, in the current main branch code, even though `-PuseLocal=true` is enabled, it's still using the version specified in `build.gradle`.

The Cron Job is not functioning as expected because it's using the default LSP4IJ version instead of the locally built plugin from the merged PRs. This is causing the intended purpose of the Cron Job to be missed.

anusreelakshmi934:
Fixes #1242


build.gradle
```diff
@@ -148,10 +148,11 @@ dependencies {
         create(providers.gradleProperty("platformType"), providers.gradleProperty("platformVersion"))
         bundledPlugins providers.gradleProperty("platformBundledPlugins").orElse("").get().split(',').toList()
 
-        def lsp4ij = providers.gradleProperty('useLocal') && providers.gradleProperty('useLocal') == 'true' ?
-                file("../lsp4ij/build/distributions/LSP4IJ/") :
-                'com.redhat.devtools.lsp4ij:' + lsp4ijVersion
-        plugins lsp4ij
+        def lsp4ijZip = file("../lsp4ij/build/distributions").listFiles()?.findAll { it.name.endsWith(".zip") }?.first()
+        providers.gradleProperty("useLocal").get() == 'true' ?
```
vaisakhkannan, line 152: I am wondering whether the condition will always evaluate to false, because we have provided a false value [here](https://github.com/OpenLiberty/liberty-tools-intellij/blob/a967e46fafa0cb2dd234529188969452b1fd61dc/gradle.properties#L1).

Can you please confirm that ? Please share the Cron Job build links that you tested with this change

vaisakhkannan (changes requested):

build.gradle
```diff
@@ -148,10 +148,11 @@ dependencies {
         create(providers.gradleProperty("platformType"), providers.gradleProperty("platformVersion"))
         bundledPlugins providers.gradleProperty("platformBundledPlugins").orElse("").get().split(',').toList()
 
-        def lsp4ij = providers.gradleProperty('useLocal') && providers.gradleProperty('useLocal') == 'true' ?
-                file("../lsp4ij/build/distributions/LSP4IJ/") :
-                'com.redhat.devtools.lsp4ij:' + lsp4ijVersion
-        plugins lsp4ij
+        def lsp4ijZip = file("../lsp4ij/build/distributions").listFiles()?.findAll { it.name.endsWith(".zip") }?.first()
```
dessina-devasia, line 151: @anusreelakshmi934 Can you please check the code format here?

anusreelakshmi934: I have formatted the code directly in IntelliJ IDE, and this is the result it provided.

build.gradle
anusreelakshmi934: By default, the value is set to `false`. However, when we run the command with `-PuseLocal=true`, it changes to `true`, allowing us to use the local LSP4IJ. I verified this by running IDE locally using `./gradlew runIdeForUiTests -PuseLocal=true`, which installed LSP4IJ `0.10.0-SNAPSHOT`, the latest version from the main branch. Without `-PuseLocal=true`, it installed version `0.8.1`, as specified in `build.gradle`.

anusreelakshmi934:
Cron Job link - https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/13025040119

dessina-devasia (approved):
Looks good to me. Thanks!

build.gradle
```diff
@@ -148,10 +148,11 @@ dependencies {
         create(providers.gradleProperty("platformType"), providers.gradleProperty("platformVersion"))
         bundledPlugins providers.gradleProperty("platformBundledPlugins").orElse("").get().split(',').toList()
 
-        def lsp4ij = providers.gradleProperty('useLocal') && providers.gradleProperty('useLocal') == 'true' ?
-                file("../lsp4ij/build/distributions/LSP4IJ/") :
-                'com.redhat.devtools.lsp4ij:' + lsp4ijVersion
-        plugins lsp4ij
```
dessina-devasia, line 154: @anusreelakshmi934 It will be better to add a comment corresponding to this change

anusreelakshmi934: I have added the comment

dessina-devasia (dismissed):
LGTM

mrglavas (changes requested):
@anusreelakshmi934 It occurred to me that there's an unzip step in our build: https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/.github/workflows/build.yaml that seems no longer necessary with these changes. We should clean that up to avoid any confusion.

image

anusreelakshmi934:
> @anusreelakshmi934 It occurred to me that there's an unzip step in our build: https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/.github/workflows/build.yaml that seems no longer necessary with these changes. We should clean that up to avoid any confusion.
> 
> image

Yes @mrglavas . I will open a ticket for this and clean up the code. Thanks

mrglavas:
> > @anusreelakshmi934 It occurred to me that there's an unzip step in our build: https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/.github/workflows/build.yaml that seems no longer necessary with these changes. We should clean that up to avoid any confusion.
> > image
> 
> Yes @mrglavas . I have cleaned up the code. Here is the Cron Job link after the change: [link](https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/13045705589/job/36395825273).
> 
> In this run, the 24.0.9 build [fails](https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/13045705589/job/36395822047#step:7:33) because it lacks the unzip step, preventing it from retrieving the `lsp4ij` folder correctly. However, the 24.0.12 build succeeds since the condition is not functioning as expected. Should we proceed with this change despite the failure in 24.0.9?

@anusreelakshmi934 I believe in our team discussion earlier today we agreed that the unzip step should be restored to allow the 24.0.9 build to run. We also agreed that a comment to document that the unzip step is required for compatibility with the 24.0.9 build.

anusreelakshmi934:
> > > @anusreelakshmi934 It occurred to me that there's an unzip step in our build: https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/.github/workflows/build.yaml that seems no longer necessary with these changes. We should clean that up to avoid any confusion.
> > > image
> > 
> > 
> > Yes @mrglavas . I have cleaned up the code. Here is the Cron Job link after the change: [link](https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/13045705589/job/36395825273).
> > In this run, the 24.0.9 build [fails](https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/13045705589/job/36395822047#step:7:33) because it lacks the unzip step, preventing it from retrieving the `lsp4ij` folder correctly. However, the 24.0.12 build succeeds since the condition is not functioning as expected. Should we proceed with this change despite the failure in 24.0.9?
> 
> @anusreelakshmi934 I believe in our team discussion earlier today we agreed that the unzip step should be restored to allow the 24.0.9 build to run. We also agreed that a comment be added to document that the unzip step is required for compatibility with the 24.0.9 build.

I have added a comment above the unzip step.

dessina-devasia (approved):
Looks good. Thanks!

mrglavas (approved):

vaisakhkannan (approved):
Looks good

PR #749, opened 2024-05-14
Issue608 - Update to use the lsp4ij from Jetbrains marketplace with 2024.1.1 intellij support

aparnamichael:
Added support for intellij version 2024.1.1

PR #1249, opened 2025-01-30
Removed  use of # from the PR number

### LINKED ISSUE #1244: Avoid use of special characters in Slack reports
In some cases the updates that appear in Slack cannot be read because the Slack application interprets special characters and displays something else instead. 

Image
In this case the "#" character could be omitted because we could still understand the message if only the issue number was shown. 

In other cases the text comes from github and the text could include "#" or "@" characters. We should search the strings for these characters and any others that Slack may use and replace the special characters with something else. It may be as simple as changing "#" to "\\#".

vaisakhkannan:
Fixes #1244 

Removed  use of Special character # from the PR number for the slack message

With the latest changes in this PR there is no issue with the text from GitHub (title of the PR), and it include '#' or '@' characters without any problems.

Screenshot 2025-02-10 at 2 50 06 PM

### DIFF
--- a/.github/workflows/pr-slack-notification.yaml
+++ b/.github/workflows/pr-slack-notification.yaml
@@ -104,13 +104,8 @@ jobs:
               if [ "$pr_draft" = "true" ]; then
                 draft_status="\n*Draft*: true"
               fi
-          
-              # Prevent Slack from auto-linking issue numbers (e.g., #778) to private channels  
-              # by adding a zero-width space (U+200B) after the `#` symbol.  
-              # This makes the text appear unchanged but prevents unintended Slack formatting.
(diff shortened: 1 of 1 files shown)


mrglavas (approved):
Looks good. Thanks.

turkeylurkey:
In the screen shot you provide in this pull request you show sample text that reads "#749." It is not replaced with "private channel" possibly because our instance of Slack does not have a channel named 749 whereas it did seem to have a channel named "777" or some other number. So this example does not seem to prove there is no problem.

Can you try a test where you use a string from a github issue you open or perhaps you use your own dummy string that reads the same as a known existing channel?

anusreelakshmi934:
@vaisakhkannan Can we completely avoid using the # sign? As Paul mentioned, there's a chance that another PR number might have a channel with that format, which could cause an issue.

anusreelakshmi934 (approved):
looks good

PR #1250, opened 2025-01-31
LTI 24.0.12: Retrieve the local Lsp4ij Path correctly

anusreelakshmi934:
In the current 24.0.12 tag, the condition for retrieving the local LSP4IJ is not working as expected. 
As a result, it defaults to version 0.8.1 specified in the `build.gradle` file. 

To address this, we have created a branch based on 24.0.12 with the necessary changes to correctly retrieve the local LSP4IJ path. This allows the cron job tag to be updated from `24.0.12` to `24.0.12-lsp4ijFix`, ensuring that the correct condition is applied and the local LSP4IJ is used as intended, enabling the cron job to run as expected.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -42,7 +42,7 @@ on:
   push:
     branches: '**'
   pull_request:
-    branches: [ main ]
+    branches: [ main, 24.0.12-lsp4ijFix ]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


dessina-devasia (approved):
Looks good. Thanks!

mrglavas (changes requested):
This PR does not have any build results attached to it. Should update the build.yaml to build this branch.

anusreelakshmi934:
> This PR does not have any build results attached to it. Should update the build.yaml to build this branch.

Thanks @mrglavas . I have added the 24.0.12-lsp4ijFix branch in the build.yaml file

mrglavas (approved):
Thanks. Looks good.

dessina-devasia (approved):

PR #1257, opened 2025-02-04
When you use the Start action do not use the custom parameters from Start...

### LINKED ISSUE #424: Regular start runs with parameters from previously run custom start
Create a run configuration using custom start action:
custom-start-1
Run with that configuration (note the "fdsa" parameter):
image
Now, run a regular start (should be no parameters):
custom-start-2
Note that the "fdsa" parameter is part of the command:
custom-start-3

### LINKED ISSUE #955: Deletion of configuration does not remove 'Start in Container' Setting when enabled

### LINKED ISSUE #977: Redesign the relationship between Liberty module and run configurations
Redesign the relationship between Liberty module and run configurations.

- [ ] Custom params and the 'run in container' option must be set in one call.
- [ ] Alternatively the Liberty module should hold a reference to the last used run config. Then we need to remove this reference if the run config is deleted. Or mark the run config as invalid when deleted.

The current work flow supported is: 
- Use `Start...` Liberty action and type some parameters and check the Run in container checkbox. Click Run.
- Stop the server.
- Use the `Start` Liberty action. You expect the server to start with the same parameters as the first step.

This behaviour is similar to Eclipse. We can discuss with the guild if this is what we really want because VS Code may not do this. Also we should examine how other IntelliJ run configs work and decide whether Liberty tools should be consistent platform to platform or if the launch behaviour on each platform should be the same for all languages or servers on that platform.

turkeylurkey:
The Liberty config run button used to try to select Start action or Start in container. This is not needed, the Start action can handle both. 
Instead of copying the parameters into the Liberty module object we just store a reference to the Liberty run config. 
The difference in handling between Start and Start in container when you choose Debug looks like a bug. This pull request fixes that. Debugging regular Liberty and Liberty in a container should be the same. 
The test case fails when run on the main branch (without this fix):
```
GradleSingleModMPProjectTest > testStartInContainerParamClearedOnConfigRemoval() FAILED
    org.opentest4j.AssertionFailedError: Server started in container when it should not have
        at org.junit.jupiter.api.AssertionUtils.fail(AssertionUtils.java:38)
        at org.junit.jupiter.api.Assertions.fail(Assertions.java:135)
        at io.openliberty.tools.intellij.it.SingleModMPProjectTestCommon.testStartInContainerParamClearedOnConfigRemoval(SingleModMPProjectTestCommon.java:1202)
        at java.base@17.0.12/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  ...
```

Fixes #977
Fixes #955 
Fixes #424

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/LibertyModule.java
+++ b/src/main/java/io/openliberty/tools/intellij/LibertyModule.java
@@ -12,6 +12,7 @@ package io.openliberty.tools.intellij;
 
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
+import io.openliberty.tools.intellij.runConfiguration.LibertyRunConfiguration;
 import io.openliberty.tools.intellij.util.BuildFile;
 import io.openliberty.tools.intellij.util.Constants;
 import org.jetbrains.plugins.terminal.ShellTerminalWidget;
@@ -26,19 +27,17 @@ public class LibertyModule {
     private Constants.ProjectType projectType;
     private String name;
     private boolean validContainerVersion;
-
-    private String customStartParams;
-    private boolean runInContainer;
-
     private boolean debugMode;
     private ShellTerminalWidget shellWidget;
+    private LibertyRunConfiguration customRunConfig;
+    private boolean useCustom;
 
     public LibertyModule(Project project) {
         this.project = project;
-        this.customStartParams = "";
-        this.runInContainer = false;
         this.debugMode = false;
         this.shellWidget = null;
+        this.customRunConfig = null;
+        this.useCustom = false;
     }
 
     public LibertyModule(Project project, VirtualFile buildFile, String name, Constants.ProjectType projectType, boolean validContainerVersion) {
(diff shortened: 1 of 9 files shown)


mrglavas (approved):
Looks good. Always nice to see green builds across the board.

anusreelakshmi934 (approved):
LGTM

PR #1267, opened 2025-02-11
Override the method suggestedName() used during config creation

### LINKED ISSUE #180: Run/Debug configuration name and project name populated up front
It would be nice to have a friendly project name at the top of the run/debug configuration. Maven mandates an artifact id in the build file and for gradle we could use the project's directory name. 

The project name could also be used as the default name for the new configuration. If there is already a config by that name another name could be generated by adding "-2" to the end of the project name. E.g. guide-getting-started, guide-getting-started-2, guide-getting-started-3 etc.

The build file is a drop down combo box and if the user selects a different build file then the project name and config name is updated.

Mock up:
Run-debug-IntelliJ-Mockup

This way the user could click Ok to save and start using the config with little work.

turkeylurkey:
The parent class already has a method to suggest a name. The new config mechanism calls this method when creating a config and also when cloning the config to detect changes while you type.

Fixes #180 

This code suggests artifact id for Maven and directory name for Gradle. 
image
image

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunConfiguration.java
+++ b/src/main/java/io/openliberty/tools/intellij/runConfiguration/LibertyRunConfiguration.java
@@ -106,6 +106,19 @@ public class LibertyRunConfiguration extends ModuleBasedConfiguration<RunConfigu
         return new LibertyRunSettingsEditor(getProject());
     }
 
+    @Override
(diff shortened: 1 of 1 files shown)


vaisakhkannan (approved):
Looks good

anusreelakshmi934 (approved):
Lgtm

mrglavas (approved):
Looks good. Thanks.

PR #1288, opened 2025-02-25
Update test case Dockerfile by switching to non root user

### LINKED ISSUE #554: Update test case Dockerfile
Investigate whether the Dockerfile used in the test cases can be updated to run with a non-root user. They do a `chown 1001` for some files, but do not switch to that user at the end of the file.

Affected files:
- https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/test/resources/projects/maven/singleModMavenMP/Dockerfile
- https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/test/resources/projects/gradle/singleModGradleMP/Dockerfile
- https://github.com/OpenLiberty/liberty-tools-intellij/blob/main/src/test/resources/projects/gradle/singleModGradleRESTNoLTXmlCfg/Dockerfile

This was identified in the static app scan.

anusreelakshmi934:
Fixes #554

### DIFF
--- a/src/test/resources/projects/gradle/singleModGradleMP/Dockerfile
+++ b/src/test/resources/projects/gradle/singleModGradleMP/Dockerfile
@@ -24,3 +24,6 @@ RUN features.sh
 COPY --chown=1001:0 build/libs/*.war /config/apps
 
 RUN configure.sh
+
+# Switch to non-root user
+USER 1001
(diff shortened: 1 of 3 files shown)


anusreelakshmi934:
Closing this PR since switching to non root user by using USER 1001 which in effect has no impact on the docker image since it is run by the user 1001 by default.

PR #1310, opened 2025-03-12
Update the Lsp4Jakarta version number to the latest

### LINKED ISSUE #1309: Support new release of LSP4Jakarta (1Q-2025)
There is one dependency that needs updating - `org.eclipse.lsp4jakarta.ls-0.2.2-jar-with-dependencies.jar`

Only one file should be affected - `build.gradle` (`implementation` and `lsp` lines)

- [ ] Smoke test to confirm no regression is introduced.

turkeylurkey:
This pull request introduces a new way to specify the version number of language server dependencies. They are needed in two locations so it makes sense to specify the number once in the `gradle.properties` (like some others are) and then to reference that value.

We also update lsp4jakarta from 0.2.2 to 0.2.3.

Fixes #1309

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -83,20 +83,20 @@ configurations {
 
 dependencies {
 
-    implementation ('org.eclipse.lsp4mp:org.eclipse.lsp4mp.ls:0.13.0') {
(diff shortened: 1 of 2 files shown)


mrglavas (approved):
Looks good. Thanks.

PR #1319, opened 2025-03-23
Increase Delay to Obtain the Diagnostic, Leading to Test Pass

### LINKED ISSUE #1204: Hover on text: 'wrong' did not trigger a pop-up window to open in testQuickFixInServerXML() and testDiagnosticInServerXML()
java.lang.RuntimeException: Hover on text: 'wrong' did not trigger a pop-up window to open 
Caused by:
        com.intellij.remoterobot.utils.WaitForConditionTimeoutException: Exceeded timeout (PT5S) for condition function (Failed to find 'Container' by '//div[@class='HeavyWeightWindow']//div[@class='JEditorPane']' in 5s) 

---
Last time bug was seen: March 6th, 5th, 3rd & 2nd 2025
Link to build: https://github.com/OpenLiberty/liberty-tools-intellij/actions/runs/13564497339/job/38075762249?pr=1296
https://github.com/OpenLiberty/liberty-tools-intellij/actions/runs/13650621388/job/38158154182 
https://github.com/OpenLiberty/liberty-tools-intellij/actions/runs/13585344984/job/37978979473

vaisakhkannan:
Part of  fix #1204

Since the corresponding tests are failing only on Windows OS, I added a delay between character entries for Windows only. I increased the iteration count from 10 to 15 to provide more delay, reduced the sleep time from 20 to 10 to avoid unnecessary delays between iterations in diagnostic test method, and increased the sleep time for the QuickFix hover data method.

I have set a delay of `200` after multiple checks because if we set it to 500 or above, we may get a timeout exception, and if we set it to 300, we may get a 'List is empty' exception, both of which cause test failures.

Screenshot 2025-03-23 at 12 20 08 PM

Screenshot 2025-03-23 at 11 35 21 AM

 So, I chose the value 200. The default value is 50, which works faster on Windows OS, but on other operating systems, it enters at a normal speed. That's why I added a condition.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -830,7 +830,6 @@ public class UIBotTestUtils {
 
         Exception error = null;
         for (int i = 0; i < 15; i++) {
-            System.out.println("//////// hover Attempt "+ i +" ------------");
             error = null;
             try {
                 // move the cursor to the origin of the editor
@@ -866,7 +865,7 @@ public class UIBotTestUtils {
                 break;
             } catch (WaitForConditionTimeoutException wftoe) {
                 error = wftoe;
-                TestUtils.sleepAndIgnoreException(5);
+                TestUtils.sleepAndIgnoreException(10);
                 // click on center of editor pane - allow hover to work on next attempt
                 editorNew.click();
             }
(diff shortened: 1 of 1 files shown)


dessina-devasia (approved):
Looks good. Thanks!

vaisakhkannan:
Build Results : https://github.com/vaisakhkannan/liberty-tools-intellij/actions/runs/14042437736

turkeylurkey (approved):
It looks like Jetbrains already encountered the problem of fast typing and added an API for that.

PR #1461, opened 2025-11-12
Updated accessibleName of files in the UI tests

### LINKED ISSUE #1431: Update the accessibleName of files in the UI tests running with the EAP version
In the EAP version (2025.3 EAP build 253.27864.23 as of October 22), the accessible names of certain files have changed. For example, **`bootstrap.properties`** is now shown as **`bootstrap.properties, properties file`**, and **`server.xml`** as **`server.xml, XML file`**. Update the tests to accommodate these changes and any other changes if present, so they continue to pass.

anusreelakshmi934:
Fixes #1431 
- Updated the XPath to use startsWith() for matching accessible names such as server.xml, bootstrap.properties etc.
- Liberty entry not visble in the Add New configuration window unless the all are collapsed. Hence introduced a Collapse All button click.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/SingleModMPLSTestCommon.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2023, 2025 IBM Corporation.
+ * Copyright (c) 2023, 2024 IBM Corporation.
  *
  * This program and the accompanying materials are made available under the
  * terms of the Eclipse Public License v. 2.0 which is available at


src/test/java/io/openliberty/tools/intellij/it/SingleModLibertyLSTestCommon.java
```diff
@@ -86,7 +86,7 @@ public void testServerXMLFeatureHover() {
         String hoverExpectedOutcome = "This feature provides support for the MicroProfile Health specification.";
 
         //mover cursor to hover point
-        UIBotTestUtils.hoverInAppServerCfgFile(remoteRobot, testHoverTarget, "server.xml", UIBotTestUtils.PopupType.DOCUMENTATION);
+        UIBotTestUtils.hoverInAppServerCfgFile(remoteRobot, testHoverTarget, "server.xml, XML file", UIBotTestUtils.PopupType.DOCUMENTATION);
```
TrevCraw: These accessible names should be constants. It eliminates potential human error having to write this or copy paste multiple times. It also helps when changes like this are required - only one spot in the code needs to be updated. It might be worth setting that up now.

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
```diff
@@ -355,20 +355,27 @@ protected static void closeProjectView() {
...
     public void testOpenBuildFileActionUsingPopUpMenu() {
         shouldCleanupTerminal = false;
-        String editorTabName = getBuildFileName() + " (" + getSmMPProjectName() + ")";
+        String editorTabNameWithProject = getBuildFileName() + " (" + getSmMPProjectName() + ")";
+        String editorTabNameSimple = getBuildFileName();
 
         // Close the editor tab if it was previously opened.
-        UIBotTestUtils.closeFileEditorTab(remoteRobot, editorTabName, "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, editorTabNameWithProject, "5");
+        UIBotTestUtils.closeFileEditorTab(remoteRobot, editorTabNameSimple, "5");
```
TrevCraw: I am guessing that the reason the `closeFileEditorTab` method is called twice is to make sure it catches both types of projects? Even though only one can be correct at a time?

It would be cleaner if we could do some sort of if statement here. Or better yet, figure out some logic where the child test classes implement the difference (i.e., make a super call passing a variable based on the type of project it is). Or the child class defines "editorTabName" and then that is used in the parent class. Something along those lines.

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
```diff
@@ -355,20 +355,27 @@ protected static void closeProjectView() {
...
 
         // Verify that build file tab is opened.
-        Assertions.assertNotNull(UIBotTestUtils.getEditorTabCloseButton(remoteRobot, editorTabName, "10"),
-                "Editor tab with the name of " + editorTabName + " could not be found.");
+        ComponentFixture editorTabCloseButton = UIBotTestUtils.getEditorTabCloseButton(remoteRobot, editorTabNameWithProject, "10");
+        if (editorTabCloseButton == null) {
+            editorTabCloseButton = UIBotTestUtils.getEditorTabCloseButton(remoteRobot, editorTabNameSimple, "10");
+        }
+
+        Assertions.assertNotNull(editorTabCloseButton, "Editor tab with the name of " + editorTabNameWithProject + " or " + editorTabNameSimple + " could not be found.");
```
TrevCraw: Same comment as above. Maybe child classes can help simplify this code?

TrevCraw (changes requested):

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
anusreelakshmi934: Hi @TrevCraw . This change isn’t needed. During my earlier testing, I observed that `editorTabName` appeared as `getBuildFileName()` for projects without spaces, and as `getBuildFileName() + " (" + getSmMPProjectName() + ")"` for projects with spaces. That’s why I initially updated the code to handle both cases.

However, it turns out the difference I saw was because the project wasn’t fully loaded in the explorer in my local environment, as shown in the screenshot. 
Screenshot 2025-11-13 at 3 51 35 PM

Once the project was properly loaded, the editor tab name displayed consistently.
So no updates are required to the editor tab name at this point. Hence I have reverted the change

src/test/java/io/openliberty/tools/intellij/it/SingleModMPProjectTestCommon.java
anusreelakshmi934: Please see my comment [here](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1461#discussion_r2523060752)

PR #1481, opened 2025-11-26
Fix test failure in testCreateRunConfigAction()

### LINKED ISSUE #1479: Investigate the test failure in testCreateRunConfigAction()
While running the test automation, we can see that testCreateRunConfigAction fails . 
Here is the build link - https://github.com/OpenLiberty/liberty-tools-intellij/actions/runs/19626340236/job/56260278043?pr=1478#step:10:1146. Despite rerunning twice, it still fails. Find out the cause and resolve the issue.

anusreelakshmi934:
Fixes #1479 

- Added waitForIndexing before each test, so that test starts after indexing is complete and it doesn't fail to find 'Combobox' 

I ran the testCreateRunConfigAction test multiple times with this fix, and it passed consistently. Sharing the build reference link that includes this change: https://github.com/anusreelakshmi934/liberty-tools-intellij/actions/runs/19673130379.


dessina-devasia (approved):
Looks good to me! Thanks!

PR #1482, opened 2025-11-26
Allow the “bash requesting screen access” popup in macOs latest

### LINKED ISSUE #1451: UI tests blocked by “bash requesting screen access” in Mac OS 15 onwards
When the tests start, the following pop-ups appear and block execution:

Image

Because of these dialogs, the UI tests hang indefinitely since they require user interaction.

anusreelakshmi934:
Fixes #1451 

The fix is to add a test that runs first in each class, ensuring the popup is brought into focus and the Allow button is clicked before any other tests run. Added the new AllowPopupTest test method in a base class that each of the classes extend.

### DIFF
--- a/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
+++ b/src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
@@ -442,7 +442,7 @@ public class UIBotTestUtils {
      */
     public static void openProjectView(RemoteRobot remoteRobot) {
         // maximize windows os intellij ide to avoid failures accessing menu
-        if (remoteRobot.isWin()) {
+        if (remoteRobot.isWin() || remoteRobot.isMac()) {
             maximizeWindow(remoteRobot);
         }
         TestUtils.printTrace(TestUtils.TraceSevLevel.INFO, "UIBotTestUtils.openProjectView Entry");
@@ -2855,7 +2855,12 @@ public class UIBotTestUtils {
     public static void maximizeWindow(RemoteRobot remoteRobot) {
         if (!isFullScreen(remoteRobot)) {
             Keyboard keyboard = new Keyboard(remoteRobot);
-            keyboard.hotKey(VK_WINDOWS, VK_UP);
+            if (remoteRobot.isWin()) {
+                keyboard.hotKey(VK_WINDOWS, VK_UP);
+            } else {
+                // macOS fullscreen shortcut: Command + Control + F
+                keyboard.hotKey(VK_META, VK_CONTROL, VK_F);
+            }
             keyboard.enter();
         }
     }


TrevCraw (commented):
Hi @anusreelakshmi934 , I know this PR is still in draft mode, I just wanted to give some early feedback.

1. First off, this looks really good. It's awesome that you found a solution for this.
2. Did you write the AppleScript code yourself or borrow from another source? If borrowed, it's important we reference where we got it from.
3. To avoid repeating the new `AllowPopupTest` test method for each of the different test classes, it would be good to create a base class that each of these classes could extend which contains the new `AllowPopupTest` test method. I see there are some variations in some of the `AllowPopupTest` methods to bring the build file into focus. Can that code be reused and just go in the base class? Maybe there is a check that can be used to see if the project is Gradle before running those steps?
4. Can you look into the test failures to verify that they are not related to these changes?

src/test/java/io/openliberty/tools/intellij/it/MacOSAllowPopupTest.java
```diff
@@ -0,0 +1,42 @@
...
+package io.openliberty.tools.intellij.it;
+
+import com.intellij.remoterobot.RemoteRobot;
+import com.automation.remarks.junit5.Video;
+import org.junit.jupiter.api.*;
+import org.junit.jupiter.api.condition.EnabledOnOs;
+import org.junit.jupiter.api.condition.OS;
+
+
+public class MacOSAllowPopupTest {
```
TrevCraw: Discussed with @turkeylurkey . We'd like to suggest renaming the class to "BaseOSUtilities". It seems strange to have our common test classes extending a Mac specific class. This class could then be added to in the future if other "tests" or setup becomes necessary for any OS.

src/test/java/io/openliberty/tools/intellij/it/UIBotTestUtils.java
```diff
@@ -2939,4 +2939,45 @@ public static void clickOnLoad(RemoteRobot remoteRobot) {
...
+        if (!remoteRobot.isMac()) {
+            return; // Only applicable to macOS
+        }
+        TestUtils.printTrace(TestUtils.TraceSevLevel.INFO, "Handling macOS permission popup...");
+
+        // Wait for the permission popup to appear
+        TestUtils.sleepAndIgnoreException(10);
+        // Execute AppleScript to click the "Allow" button
+        try {
+            String appleScript =
```
TrevCraw: It would be good to cite the source that was used as reference for this "Apple script". You can mention that the appleScript is based off an example provided at https://smartwatermelon.medium.com/automating-macos-security-dialogs-a-tale-of-yak-shaving-and-applescript-759300d6fba9. It would also be good to point to official documentation, like the one you shared here: https://developer.apple.com/library/archive/documentation/LanguagesUtilities/Conceptual/MacAutomationScriptingGuide/AutomatetheUserInterface.html - just in case a developer in the future needs to make changes.

anusreelakshmi934: I have added these references.

src/test/java/io/openliberty/tools/intellij/it/SingleModJakartaLSTestCommon.java
```diff
@@ -21,7 +21,8 @@
 
 import static com.intellij.remoterobot.utils.RepeatUtilsKt.waitForIgnoringError;
 
-public abstract class SingleModJakartaLSTestCommon {
+@TestMethodOrder(MethodOrderer.OrderAnnotation.class)
```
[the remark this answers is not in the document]
anusreelakshmi934: Yes. I have added it to just base class and removed it from others. It works as expected.

PR #1492, opened 2025-12-01
Implementation to capture log from LSP console

### LINKED ISSUE #1491: Implement an approach to capture log from LSP console
Implement test automation to enable trace logs at debug level, capture the LSP console, and write it to a log file that can also be uploaded to GHA

dessina-devasia:
Fixes #1491 

Implemented test automation to enable trace logs at debug level, capture the LSP console, and write it to a log file that can also be uploaded to GHA


angelozerr:
Your PR looks like very complex to capture lsp traces from console. If you need this feature please create an  issue to lsp4ij

PR #1535, opened 2026-02-25
Fixes #1534 Sync diagnostics for JSONP Object Builder Null key from lsp4jakarta to IntelliJ

### LINKED ISSUE #1534: Sync diagnostics add diagnostics for JSONP Object Builder Null key from lsp4jakarta to IntelliJ
lsp4jakarta issue: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/639
lsp4jakarta PR: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/817

archana-1924:
Fixes #1534 

lsp4jakarta issue: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/639
lsp4jakarta PR: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/817

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/jsonp/JsonpDiagnosticCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/jsonp/JsonpDiagnosticCollector.java
@@ -45,11 +45,12 @@ public class JsonpDiagnosticCollector extends AbstractDiagnosticsCollector {
         List<PsiMethodCallExpression> createPointerInvocations = new ArrayList<>();
         //Used to get the list of method invocations for JsonObjectBuilder add methods
         List<PsiMethodCallExpression> createObjectBuilderMethodInvocations = new ArrayList<>();
+        String createPointerFQName = JsonpConstants.JSON_FQ_NAME + "." + JsonpConstants.CREATE_POINTER;
         for (PsiMethodCallExpression mi : allMethodInvocations) {
-            if (isMatchedJsonCreatePointer(mi)) {
+            if (isMatchedMethodFQName(mi,JsonpConstants.EXPRESSION_COUNT_CREATE_POINTER, createPointerFQName)) {
                 createPointerInvocations.add(mi);
             }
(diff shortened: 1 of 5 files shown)


Joseph-Bineesh (approved):
LGTM

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/jsonp/JsonpDiagnosticCollector.java
```diff
@@ -59,6 +64,34 @@ public void collectDiagnostics(PsiJavaFile unit, List<Diagnostic> diagnostics) {
...
+    }
+
+    /**
+     * isMatchedJsonObjectBuilder
+     * Method used to identify jakarta.json.JsonObjectBuilder.add type method invocations
+     *
+     * @param mi
+     * @return boolean
+     */
+    private boolean isMatchedJsonObjectBuilder(PsiMethodCallExpression mi) {
```
venmanyarun: this method is very similar to isMatchedJsonCreatePointer()

can we extract common logic?

venmanyarun (changes requested):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/jsonp/JsonpDiagnosticCollector.java
archana-1924: Done, it affects the current func too, so did regression.

venmanyarun (approved):

turkeylurkey (approved):
Looks good, but I have one comment.

The code that uses this technique 

```
        String fqName = declaringClass.getQualifiedName() + "." + binding.getName();
        if (Constants.JAKARTA_JSON_OBJECT_BUILDER_ADD.equals(fqName)) {
```
is kind of an antipattern. It takes organized data and homogenizes it before use. In fact, this even has a performance impact as well because we build a string and perform a rather costly comparison with it. It would be more efficient to compare the method names first and only compare the class names if required. 

Since there is already code like this you might want to open an issue to rewrite this in the future.

PR #1554, opened 2026-03-23
PR for Lsp4jakarta issue 677: Diagnostics for missing primary key in entity classes

Joseph-Bineesh:
Issue: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/677
lsp4jakarta PR: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/831

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
@@ -149,4 +149,15 @@ public class DiagnosticsUtils {
         }
         return false;
     }
+
+    public static List<PsiClass> collectSuperClasses(PsiClass psiClass) {
+        List<PsiClass> superClasses = new ArrayList<>();
+        PsiClass current = psiClass.getSuperClass();
+
+        while (current != null) {
+            superClasses.add(current);
+            current = current.getSuperClass();
+        }
+        return superClasses;
+    }
 }
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
@@ -15,6 +15,7 @@ package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.persistence;
 
 import com.intellij.psi.*;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.AbstractDiagnosticsCollector;
+import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.DiagnosticsUtils;
 import io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.Messages;
 import org.eclipse.lsp4j.Diagnostic;
 import org.eclipse.lsp4j.DiagnosticSeverity;
@@ -60,6 +61,7 @@ public class PersistenceEntityDiagnosticsCollector extends AbstractDiagnosticsCo
                     boolean hasPublicOrProtectedNoArgConstructor = false;
                     boolean hasArgConstructor = false;
                     boolean isEntityClassFinal = false;
+                    boolean hasPrimaryKey = false;
 
                     // Get the Methods of the annotated Class
                     for (PsiMethod method : type.getMethods()) {
@@ -82,6 +84,12 @@ public class PersistenceEntityDiagnosticsCollector extends AbstractDiagnosticsCo
                                     PersistenceConstants.DIAGNOSTIC_CODE_FINAL_METHODS, method.getReturnType().getInternalCanonicalText(),
                                     DiagnosticSeverity.Error));
                         }
+
+                        // Check if any method has @Id or @EmbeddedId annotation
+                        if (!hasPrimaryKey && hasPrimaryKeyAnnotation(type, method.getAnnotations())) {
+                            hasPrimaryKey = true;
+                        }
+
                     }
 
                     // Go through the instance variables and make sure no instance vars are final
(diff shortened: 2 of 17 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
```diff
@@ -207,4 +235,64 @@ private void validatePKDateTemporal(PsiJvmModifiersOwner fieldOrProperty, PsiCla
...
+
+    /**
+     * Check if the given annotations contain @Id or @EmbeddedId
+     *
+     * @param type the type context for resolving annotations
+     * @param annotations the annotations to check
+     * @return true if a primary key annotation is found
+     */
+    private boolean hasPrimaryKeyAnnotation(PsiClass type, PsiAnnotation[] annotations) {
+        for (PsiAnnotation annotation : annotations) {
```
archana-1924: Please sync up changes in lsp4jakarta

archana-1924 (changes requested):

archana-1924 (approved):
LGTM

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
```diff
@@ -149,4 +149,15 @@ public static boolean isMatchedJavaElement(PsiClass type, String javaElementName
         }
         return false;
     }
+
+    public static List<PsiClass> collectSuperClasses(PsiClass psiClass) {
+        List<PsiClass> superClasses = new ArrayList<>();
+        PsiClass current = psiClass.getSuperClass();
+
+        while (current != null) {
```
venmanyarun, line 157: I think the while should be changed as below to avoid infinite loops with object.class   while (current != null && !"java.lang.Object".equals(current.getQualifiedName())) {}

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
```diff
@@ -207,4 +236,60 @@ private void validatePKDateTemporal(PsiJvmModifiersOwner fieldOrProperty, PsiCla
...
+     *
+     * @param type the type to check
+     * @return true if a primary key is found in the hierarchy
+     */
+    private boolean hasPrimaryKeyInSuperclass(PsiClass type) {
+        List<PsiClass> hierarchySuperClasses = DiagnosticsUtils.collectSuperClasses(type);
+
+        for (PsiClass superClass : hierarchySuperClasses) {
+            if (superClass.equals(type)) {
+                continue;
```
venmanyarun: can you explain the scenario where currentClass and superClass type would be same? looks like this check is redundant.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
```diff
@@ -207,4 +236,60 @@ private void validatePKDateTemporal(PsiJvmModifiersOwner fieldOrProperty, PsiCla
...
+            if (isMappedSuperclass) {
+                // Check fields in superclass
+                for (PsiField field : superClass.getFields()) {
+                    if (hasPrimaryKeyAnnotation(superClass, field.getAnnotations())) {
+                        return true;
+                    }
+                }
+
+                // Check methods in superclass
+                for (PsiMethod method : superClass.getMethods()) {
```
venmanyarun, line 282: This code is duplicated

Lines 89-92 (for methods)
Lines 114-117 (for fields)
Lines 275-289 (in superclass checking)

can we extract this and write a separate helper method?

venmanyarun (changes requested):
shouldn't we add test for
Entity with @IdClass (composite key alternative)

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
Joseph-Bineesh: Even if it has value java.lang.Object that will end up with null in the next iteration. Because it is the root of the class hierarchy. So can you explain the infinite loop scenario?

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
Joseph-Bineesh: Lines 89-92 (for methods) - reusing the existing loop
Lines 114-117 (for fields) - reusing the existing loop
Lines 275-289 (in superclass checking) - its own implementation

If we extract to a helper method we have to loop again and validate.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
Joseph-Bineesh: Ya this is true. This only relevant when we add current class in collectSuperClasses list. This can be avoid

Joseph-Bineesh:
> shouldn't we add test for Entity with @IdClass (composite key alternative)

I think this is a missing implementation. I did check for only @Id and @EmbededID. Thanks for pointing out. I will add the cases

Joseph-Bineesh:
> > shouldn't we add test for Entity with @IdClass (composite key alternative)
> 
> I think this is a missing implementation. I did check for only @id and @EmbededID. Thanks for pointing out. I will add the cases

@venmanyarun 
@IdClass cannot be used without @Id, because the primary key is defined on the entity fields themselves. You must define primary key fields in the entity using @Id and Field names and types must match between entity and IdClass.

We have covered the diagnostics for @Id so implicitly it covered @IdClass

venmanyarun (approved):

turkeylurkey (approved):
Looks good, thank you.

PR #1620, opened 2026-06-22
ejb message driven bean MessageListener interface diagnostic and quickfix

venmanyarun:
Fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/646
lsp4jakarta sync PR:- https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/898

LTI screen recording :- 

https://github.com/user-attachments/assets/00c81512-5b54-4da3-944c-8ae07a1007d6

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/Constants.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/Constants.java
@@ -1,0 +1,29 @@
+/*******************************************************************************
+ * Copyright (c) 2026 IBM Corporation and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v. 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.ejb;
+
+/**
+ * Constants for Jakarta Enterprise Beans (EJB) diagnostics.
+ */
+public class Constants {
+
+    /* Annotations */
+    public static final String MESSAGE_DRIVEN_FQ_NAME = "jakarta.ejb.MessageDriven";
+    
+    /* Interfaces */
+    public static final String MESSAGE_LISTENER_FQ_NAME = "jakarta.jms.MessageListener";
+    
+    /* Diagnostic codes */
+    public static final String DIAGNOSTIC_SOURCE = "jakarta-ejb";
+    public static final String DIAGNOSTIC_CODE = "ImplementMessageListener";
+}
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbConstants.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbConstants.java
@@ -15,7 +15,7 @@ package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.ejb;
 /**
  * Constants for Jakarta Enterprise Beans (EJB) diagnostics.
  */
-public class Constants {
+public class EjbConstants {
 
     /* Annotations */
     public static final String MESSAGE_DRIVEN_FQ_NAME = "jakarta.ejb.MessageDriven";
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbConstants.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbConstants.java
@@ -33,6 +33,6 @@ public class EjbConstants {
 
     /* Diagnostic codes */
     public static final String DIAGNOSTIC_SOURCE = "jakarta-ejb";
-    public static final String DIAGNOSTIC_CODE = "ImplementMessageListener";
+    public static final String DIAGNOSTIC_CODE_MISSING_MESSAGELISTENER = "ImplementMessageListener";
     public static final String DIAGNOSTIC_CODE_MISSING_PUBLIC_CONSTRUCTOR = "MissingPublicNoArgConstructor";
 }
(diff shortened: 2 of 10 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/Constants.java
```diff
@@ -0,0 +1,29 @@
...
+ *
+ * Contributors:
+ *     IBM Corporation - initial API and implementation
+ *******************************************************************************/
+package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij.ejb;
+
+/**
+ * Constants for Jakarta Enterprise Beans (EJB) diagnostics.
+ */
+public class Constants {
```
sajeerzeji: Every other module uses <Feature>Constants.java (e.g., ServletConstants.java, ManagedBeanConstants.java, BeanValidationConstants.java, JsonbConstants.java), can we rename this one as well?

src/test/java/io/openliberty/tools/intellij/lsp4jakarta/it/ejb/EjbMessageDrivenTest.java
```diff
@@ -0,0 +1,162 @@
...
+import java.io.File;
+import java.util.Arrays;
+
+import com.intellij.openapi.module.Module;
+import com.intellij.openapi.module.ModuleUtilCore;
+import com.intellij.openapi.vfs.LocalFileSystem;
+import com.intellij.openapi.vfs.VfsUtilCore;
+import com.intellij.openapi.vfs.VirtualFile;
+import io.openliberty.tools.intellij.lsp4jakarta.it.core.BaseJakartaTest;
+import io.openliberty.tools.intellij.lsp4jakarta.it.core.JakartaForJavaAssert;
```
sajeerzeji: The majority of existing tests (e.g., BeanValidationTest, ManagedBeanTest, JakartaWebSocketTest, etc.) does static import for JakartaForJavaAssert.*, shouldn't we be consistent here too?

venmanyarun: Done.. corrected in lsp4jakarta PR as well

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/Constants.java
venmanyarun: Done.. corrected in lsp4jakarta PR as well

sajeerzeji (dismissed):
LGTM

sajeerzeji (dismissed):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbMessageDrivenDiagnosticsCollector.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+    }
+
+    /**
+     * Checks if the given class implements jakarta.jms.MessageListener interface.
+     * 
+     * @param type the class to check
+     * @return true if the class implements MessageListener, false otherwise
+     */
+    private boolean implementsMessageListener(PsiClass type) {
```
Rejoice472, line 90: Could you please check if it’s possible to use the existing API DiagnosticsUtils.inheritsFrom?

Rejoice472 (changes requested):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbMessageDrivenDiagnosticsCollector.java
venmanyarun: yes changed to use existing API DiagnosticsUtils.inheritsFrom instead of 

return type.isInheritor(messageListenerClass, true);

The other logic for null-resolution guard is still needed

sajeerzeji (dismissed):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbMessageDrivenDiagnosticsCollector.java
venmanyarun: We cannot use DiagnosticsUtils.inheritsFrom(type, MESSAGE_LISTENER_FQ_NAME);

There's one behavioural difference to note: when MessageListener is not on the classpath, the old code returned true (assume valid, no false-positives), but inheritsFrom(PsiClass, String) delegates to inheritsFrom(PsiClass, PsiClass) which returns false when superClass == null — meaning it would fire a false-positive diagnostic. The fix needs to preserve the original guard:

Rejoice472 (dismissed):

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbConstants.java
```diff
@@ -14,19 +14,25 @@
...
+    public static final String MESSAGE_LISTENER_FQ_NAME = "jakarta.jms.MessageListener";
     public static final String[] SESSION_BEAN_ANNOTATIONS = {
             STATELESS_FQ_NAME,
             STATEFUL_FQ_NAME,
             SINGLETON_FQ_NAME
     };
 
+    /* Diagnostic codes */
     public static final String DIAGNOSTIC_SOURCE = "jakarta-ejb";
+    public static final String DIAGNOSTIC_CODE = "ImplementMessageListener";
```
turkeylurkey: I think the name "DIAGNOSTIC_CODE" could be better. Here we have a list of named constants used for similar purposes and which are differentiated by their names. We see the existing constant is named "DIAGNOSTIC_CODExxx" where xxx suggests what the constant is used for. But now we add a new constant with just a very generic name even though it is used for one specific diagnostic. I'd suggest to stay with the system of the name indicating what the constant is for. That is, change "DIAGNOSTIC_CODE" to something that suggests what kind of diagnostic it is for.

turkeylurkey (changes requested):
One improvement noted.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbConstants.java
venmanyarun: Done

Rejoice472 (approved):

sajeerzeji (approved):

turkeylurkey (approved):
Looks good!

PR #1628, opened 2026-06-25
Session bean class must not declare a finalize() method Diagnostic and Quickfix

venmanyarun:
Fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/634
Lsp4Jakarta PR :- https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/904

Recording :- 

https://github.com/user-attachments/assets/4bc111d3-8106-4b5c-afa8-ffd2a71cfed6

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbDiagnosticsCollector.java
@@ -52,6 +52,7 @@ public class EjbDiagnosticsCollector extends AbstractDiagnosticsCollector {
 
             if (!sessionBeanAnnotations.isEmpty()) {
                 validateSessionBeanConstructor(type, unit, diagnostics);
+                validateSessionBeanFinalizeMethod(type, unit, diagnostics);
             }
         }
     }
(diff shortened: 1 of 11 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbDiagnosticsCollector.java
```diff
@@ -98,4 +99,31 @@ private void validateSessionBeanConstructor(PsiClass type, PsiJavaFile unit, Lis
...
+     * @param type the class to validate
+     * @param unit the compilation unit
+     * @param diagnostics the list to add diagnostics to
+     */
+    private void validateSessionBeanFinalizeMethod(PsiClass type, PsiJavaFile unit, List<Diagnostic> diagnostics) {
+        // Check all methods in the class
+        for (PsiMethod method : type.getMethods()) {
+            // Check if this is the finalize() method
+            if (FINALIZE_METHOD_NAME.equals(method.getName()) &&
+                    method.getParameterList().getParametersCount() == 0) {
```
Joseph-Bineesh, line 119: Should we consider all the method signature?

venmanyarun: No — only finalize() with zero parameters can override Object.finalize(), which is what the spec prohibits. A finalize(...) with arguments would be an overload, not an override, and is not restricted by the spec. The check is intentionally minimal.

sajeerzeji (dismissed):
LGTM

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/ejb/EjbDiagnosticsCollector.java
Joseph-Bineesh: Parameter you already covered. I meant return type of method.

public class MyClass {
protected int finalize() throws Throwable {
return 0;
}
}
Will it trigger diagnostics for this?

Joseph-Bineesh (dismissed):
LGTM

sajeerzeji (approved):

Joseph-Bineesh (approved):
Stale review - reapproval

turkeylurkey (approved):
👍

Issue #1666, opened 2026-07-21
LibertyDevStopAction triggers EDT threading assertion


dessina-devasia:
Invoking the Liberty Dev Stop action throws an IntelliJ threading assertion

```
com.intellij.openapi.diagnostic.RuntimeExceptionWithAttachments: Access from Event Dispatch Thread (EDT) is not allowed; see https://jb.gg/ij-platform-threading for details
Current thread: Thread[#63,AWT-EventQueue-0,6,main] 298630442 (EventQueue.isDispatchThread()=true)
SystemEventQueueThread: (same)
	at com.intellij.util.concurrency.ThreadingAssertions.createThreadRequirementException(ThreadingAssertions.java:281)
	at com.intellij.util.concurrency.ThreadingAssertions.softAssertBackgroundThread(ThreadingAssertions.java:113)
	at org.jetbrains.plugins.terminal.TerminalUtil.hasRunningCommands(TerminalUtil.java:48)
	at org.jetbrains.plugins.terminal.ShellTerminalWidget.hasRunningCommands(ShellTerminalWidget.java:218)
	at io.openliberty.tools.intellij.actions.LibertyGeneralAction.getTerminalWidgetWithFocus(LibertyGeneralAction.java:180)
	at io.openliberty.tools.intellij.actions.LibertyDevStopAction.executeLibertyAction(LibertyDevStopAction.java:34)
	at io.openliberty.tools.intellij.actions.LibertyGeneralAction.actionPerformed(LibertyGeneralAction.java:122)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.doPerformActionOrShowPopup(ActionUtil.kt:463)
	at com.intellij.openapi.actionSystem.ex.ActionUtil.performAction$lambda$0(ActionUtil.kt:423)
	at com.intellij.openapi.actionSystem.impl.ActionManagerImpl.performWithActionCallbacks$lambda$0$0(ActionManagerImpl.kt:1234)
	at com.intellij.openapi.application.WriteIntentReadAction.lambda$run$0(WriteIntentReadAction.java:24)
	at com.intellij.openapi.application.impl.AppImplKt$rethrowCheckedExceptions$2.invoke(appImpl.kt:126)
	at com.intellij.platform.locking.impl.NestedLocksThreadingSupport.runWriteIntentReadAction(NestedLocksThreadingSupport.kt:705)
[...]

PR #1687, opened 2026-08-04
Fix EDT threading assertions thrown by LibertyDevStopAction and LibertyDevRunTestsAction from IntelliJ 2026.1

dessina-devasia:
Temporary workaround for #1666 

`LibertyGeneralAction.getTerminalWidgetWithFocus()` called `widget.hasRunningCommands()` directly on the EDT, which is the thread all IntelliJ actions run on.

Only `hasRunningCommands()` is moved off the EDT - everything else (terminal widget creation via `createShellWidget`, error notifications) stays on the EDT where those APIs require it.

A private `computeOffEdt(Supplier<T>)` helper is added to `LibertyGeneralAction` that submits the call to a pooled thread and blocks for the result.

**Files changed**
`src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java`
- Added `computeOffEdt(Supplier<T>)` helper
- Replaced direct `widget.hasRunningCommands()` call with `computeOffEdt(widget::hasRunningCommands)`

Tested with different intelliJ versions and no EDT exception observed.

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
+++ b/src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
@@ -15,6 +15,7 @@ import com.intellij.notification.Notifications;
 import com.intellij.openapi.actionSystem.ActionUpdateThread;
 import com.intellij.openapi.actionSystem.AnAction;
 import com.intellij.openapi.actionSystem.AnActionEvent;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.diagnostic.Logger;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.vfs.VirtualFile;
@@ -30,7 +31,10 @@ import org.jetbrains.plugins.terminal.TerminalToolWindowManager;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Future;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 public abstract class LibertyGeneralAction extends AnAction {
     protected static final Logger LOGGER = Logger.getInstance(LibertyGeneralAction.class);
@@ -177,7 +181,8 @@ public abstract class LibertyGeneralAction extends AnAction {
         LibertyProjectUtil.setFocusToWidget(project, existingWidget);
 
         // Shows error for actions where terminal widget does not exist or action requires a terminal to already exist and expects "Start" to be running
-        if (widget == null || (!createWidget && !widget.hasRunningCommands())) {
+        // hasRunningCommands() must not be called from the EDT (asserted since IntelliJ 2026.1)
+        if (widget == null || (!createWidget && !computeOffEdt(widget::hasRunningCommands))) {
             String msg;
             if (createWidget) {
                 msg = LocalizedResourceUtil.getMessage("liberty.terminal.cannot.resolve", actionCmd, project.getName());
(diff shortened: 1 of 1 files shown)


anusreelakshmi934 (dismissed):
Looks good

mrglavas (commented):
@dessina-devasia @TrevCraw When we discussed this PR today we agreed that it does not resolve the concern about a UI freeze because `Future.get()` blocks the EDT. We also agreed to move forward with this PR as a tactical solution but then to open a new issue for further investigation.

I have consulted with Bob on a possible alternative solution and it gave the following suggestion:

> The cleanest fix that avoids blocking the EDT entirely would restructure getTerminalWidgetWithFocus so that the hasRunningCommands() check — and everything that depends on its result — runs on a pooled thread, with any UI (notifications, widget creation) posted back via invokeLater. Since createShellWidget and error notifications also touch UI, a ReadAction.nonBlocking() / AppExecutorUtil pattern would work well here. That's a larger refactor but eliminates the freeze risk.
> 
> If a minimal fix is preferred, the executeOnPooledThread + future.get() approach is technically functional for a fast check like this, but it should be documented with a comment explaining why blocking is acceptable here (e.g., "fast flag check, no locks held"), and the ExecutionException and InterruptedException handling should be tightened as noted above.

Not sure if what Bob is suggesting is achievable but it is something to consider when this is revisited.

Some more insight from Bob on the issue with blocking the EDT. Note the additional concern about deadlocks that this introduces in general.

> Calling Future.get() from the EDT blocks the AWT event pump for as long as the pooled-thread task takes. IntelliJ's own threading guidelines (jb.gg/ij-platform-threading) explicitly state that the EDT must never be blocked. This can cause:
> - UI freezes visible to the user (a spinner/freeze while hasRunningCommands() resolves)
> - Deadlocks if the pooled thread ever needs to acquire a read/write lock that the EDT holds

Check with JetBrains on the concerns which got them to add the threading assertion for `hasRunningCommands()`.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -204,4 +209,20 @@ protected ShellTerminalWidget getTerminalWidgetWithFocus(boolean createWidget, P
...
+     * Runs the given supplier on a pooled thread and blocks for the result.
+     * Use this to call APIs that assert they must not be called from the EDT.
+     */
+    private static <T> T computeOffEdt(Supplier<T> supplier) {
+        Future<T> future = ApplicationManager.getApplication().executeOnPooledThread(supplier::get);
+        try {
+            return future.get();
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
```
mrglavas: @dessina-devasia The receiver of this exception will see a generic `RuntimeException` with no hint of the interrupt. Consider wrapping with `ProcessCanceledException` instead as that is IntelliJ's way of signalling cancellation and will be handled gracefully by the platform should this ever occur.

```suggestion
            throw new ProcessCanceledException(e);
```

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
```diff
@@ -204,4 +209,20 @@ protected ShellTerminalWidget getTerminalWidgetWithFocus(boolean createWidget, P
...
+        } catch (ExecutionException e) {
+            throw new RuntimeException(e.getCause());
```
mrglavas: @dessina-devasia `ExecutionException`'s cause is swallowed into an untyped `RuntimeException`.

This would at least preserve the type for specific `RuntimeException`s.
```suggestion
            Throwable cause = e.getCause();
            if (cause instanceof RuntimeException) throw (RuntimeException) cause;
            throw new RuntimeException(cause != null ? cause : e);
```

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
mrglavas, line 1: @dessina-devasia I recall that you mentioned that you used Bob for assistance in producing this PR. That should be included as a comment in the code.

mrglavas (changes requested):
@dessina-devasia See review comments.

src/main/java/io/openliberty/tools/intellij/actions/LibertyGeneralAction.java
dessina-devasia: Added comment. Thanks!

mrglavas (approved):
@dessina-devasia Looks good now. Thanks.

TrevCraw (approved):

anusreelakshmi934 (approved):
Looks good. Thanks

PR #1692, opened 2026-08-06
Changed build script

archana-1924:
Build script changes are as below:

- [x] Triggers only on push/pull_request to the lsp4jakarta branch
- [x] Runs only the LSP4Jakarta-Unit test group on ubuntu-latest (it currently also only runs on linux, disabled for mac and windows)
- [x] Removes the entire fetch_merge_commit_sha_from_lsp4ij_PR job


TrevCraw:
I'm not sure it makes sense to put these changes in the LSP4Jakarta integration branch, since we will have to revert the changes before merging to the main branch.

Would it be possible to have two sets of test groups? One set runs on PRs to main and contains all the test groups and the other runs on PRs to the LSP4Jakarta integration branch and contains a subset of test groups. A change like that can be committed to the main branch.

TrevCraw:
What was the reasoning behind removing the `fetch_merge_commit_sha_from_lsp4ij_PR` job? That should only run under certain conditions, which I do not believe will occur for regular PRs to the LSP4Jakarta integration branch.
cc @anusreelakshmi934

TrevCraw:
This is the list of test groups that should be run for LSP4Jakarta PRs:

- LSP4Jakarta-Unit
- Gradle-Language-Server
- Gradle-MP-Language-Server
- Gradle-Jakarta-Language-Server

.github/workflows/pr-slack-notification.yaml
```diff
@@ -7,6 +7,7 @@ on:
 
 jobs:
   fetch_all_pull_requests_and_notify_using_condition_and_cron:
+    if: github.ref == 'refs/heads/main'
```
anusreelakshmi934: Hi @archana-1924 . I have done a change to restrict this job to main openliberty repo, skip on forks. Please see the PR - https://github.com/OpenLiberty/liberty-tools-intellij/pull/1698
cc @TrevCraw

PR #1698, opened 2026-08-07
Disable all jobs from running on forks.

### LINKED ISSUE #1697: Disable Unnecessary GitHub Actions Workflows and Restrict CI Execution
Tasks

- [ ] Ensure the LSP4IJ testing cron job is completely disabled and not running on any branch or any fork.

- [ ] Disable all jobs from running on forks.

- [ ] Disable the LSP4IJ monitoring cron job from running on any branch other than main. This job is running on all active dev branches. We need to prevent that.

- [ ] Look into the possibility of preventing jobs from running on new commits to PR branches. And look into how we can manually kick off jobs on a PR. We should only run a new set of jobs after all updates are complete.

anusreelakshmi934:
Fixes #1697 

- This PR disable all jobs from running on forks. Now it only runs on PRs
- Skip this pr-slack-notification.yaml job running on forked repositories
- The trigger is restricted to 'types: [labeled]' so the workflow only starts when a label is explicitly applied , not on every push. When the PR is ready, apply the 'run-ci' label from the PR sidebar. This fires the 'pull_request: labeled' event and starts the build.

How it works:
1. Open a PR and push all your commits without CI running automatically.
2. When the PR is ready, apply the 'run-ci' label from the PR sidebar. This fires the 'pull_request: labeled' event and starts the build.
3. To re-trigger after further commits, remove and re-add the 'run-ci' label. GitHub only fires the event on the label transition (unlabeled → labeled), so the label must be toggled — not just left in place.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -47,6 +47,8 @@ on:
   push:
     # Restricted to main to avoid triggering workflows on forks.
     branches: [ main ]
+  # Runs only when the 'run-ci' label is applied to the PR, preventing CI from triggering on every push.
+  # To re-trigger after new commits, remove and re-add the label.
   pull_request:
     types: [ labeled ]
     branches: [ main ]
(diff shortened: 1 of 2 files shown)


TrevCraw (approved):

vaisakhkannan (approved):
changes looks good, Thanks

PR #1700, opened 2026-08-11
Changed build script

archana-1924:
Build script changes are as below:

- [x] Runs only the test-group builds for jakarta PRs when run-lsp4jakarta-it label is added:
            > LSP4Jakarta-Unit
            > Gradle-Language-Server
            > Gradle-MP-Language-Server
            > Gradle-Jakarta-Language-Server
- [x] Added a build section for build_lsp4jakarta which mimics the current build with subset of test-groups. This approach was adopted as dynamic setting of test-groups by manipulating matrix and fromJson() was not being recognized by Github during runtime. This is a current limitation due to which we had to add a separate build section for lsp4jakarta.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,12 +51,12 @@ on:
   # To re-trigger after new commits, remove and re-add the label.
   pull_request:
     types: [ labeled ]
-    branches: [ main ]
+    branches: [ main, 'lsp4jakarta-*-integration*', 'lsp4jakarta-*-release*' ]
 
 jobs:
   fetch_merge_commit_sha_from_lsp4ij_PR:
-    # Only run when the 'run-ci' or 'run-lsp4jakarta' label is applied to the PR; ignored for push and workflow_call/workflow_dispatch triggers.
-    if: github.event_name != 'pull_request' || github.event.label.name == 'run-ci' || github.event.label.name == 'run-lsp4jakarta'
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
venmanyarun: I think we need to configure lsp4jakarta branches here..

.github/workflows/build.yaml
```diff
@@ -55,8 +55,8 @@ on:
 
 jobs:
   fetch_merge_commit_sha_from_lsp4ij_PR:
-    # Only run when the 'run-ci' label is applied to the PR; ignored for push and workflow_call/workflow_dispatch triggers.
-    if: github.event_name != 'pull_request' || github.event.label.name == 'run-ci'
+    # Only run when the 'run-ci' or 'run-lsp4jakarta' label is applied to the PR; ignored for push and workflow_call/workflow_dispatch triggers.
```
venmanyarun: should we change to run-lsp4jakarta-it ?

venmanyarun (changes requested):

venmanyarun (approved):

anusreelakshmi934 (approved):
Looks good. Thanks

TrevCraw (approved):

TrevCraw:
@archana-1924 , this looks good. Do you know why many of the jobs are being cancelled?

venmanyarun:
> @archana-1924 , this looks good. Do you know why many of the jobs are being cancelled?

we just cancelled manually after testing whether build is triggering for the right label

PR #691, opened 2024-03-18
Add release notes for 24.0.3

turkeylurkey:
From the note in box which was reviewed.

Part of #679

PR #1706, opened 2026-08-17
Sync Fixes #691 Diagnostics for Lifecycle callback interceptor methods declared in an interceptor class or superclass of an interceptor class must have one of the signatures void (InvocationContext) or Object (InvocationContext).

archana-1924:
lsp4jakarta issue: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/691
lsp4jakarta PR: https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/957

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/interceptor/InterceptorDiagnosticsParticipant.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/interceptor/InterceptorDiagnosticsParticipant.java
@@ -100,6 +100,8 @@ public class InterceptorDiagnosticsParticipant extends AbstractDiagnosticsCollec
 						addInvalidModifierDiagnostic(method, unit, diagnostics, interceptorTypeMethodAnnotations,
 								messageKey, DIAGNOSTIC_CODE_INTERCEPTOR_STATIC, severity);
 					}
+					// Validate lifecycle callback method signatures
+					validateLifecycleCallbackMethodSignature(type, method, unit, diagnostics);
 				}
 				// Check for duplicate interceptor method annotations
 				validateDuplicateInterceptorMethods(methodsByAnnotationType, unit, diagnostics);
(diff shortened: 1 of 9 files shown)


archana-1924:
https://github.com/user-attachments/assets/f9c634ba-c9ee-4dbd-9867-3ad333b4d9af

src/test/resources/projects/maven/jakarta-sample/src/main/java/io/openliberty/sample/jakarta/interceptor/NonLifecycleInterceptorAnnotations.java
```diff
@@ -0,0 +1,28 @@
...
+import jakarta.interceptor.InvocationContext;
+
+/**
+ * Test resource: @AroundInvoke and @AroundTimeout with non-void/non-Object return types
+ * must NOT trigger InvalidLifecycleCallbackInterceptorMethodSignature — they are not
+ * lifecycle callbacks.
+ */
+@Monitored
+@Interceptor
+public class NonLifecycleInterceptorAnnotations {
```
sajeerzeji, line 15: Not a blocker, but this test resource file is different from what we have in the lsp4jakarta(The annotations used in LTI and the string return type in lsp4jakarta). If it should be differ like this, then it is fine.

archana-1924: It doesn't affect our scenarios as this is a positive scenario and doesn't trigger any diagnostics either way. But syncing it up with LSP.

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/interceptor/InterceptorDiagnosticsParticipant.java
```diff
@@ -21,6 +21,8 @@
 import java.util.HashMap;
 import java.util.stream.Collectors;
 import java.util.List;
+import com.intellij.psi.search.GlobalSearchScope;
```
Rejoice472: GlobalSearchScope import is redundant — it's already inherited from AbstractDiagnosticsCollector

PR #1725, opened 2026-08-24
Implements Jakarta Persistence 3.0 §2.9 bidirectional relationship rules

venmanyarun:
Fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/726

lsp4jakarta PR :- https://github.com/eclipse-lsp4jakarta/lsp4jakarta/pull/965

Implements Jakarta Persistence 3.0 §2.9 bidirectional relationship rules via a new PersistenceBidirectionalDiagnosticsCollector with cross-file (module-wide) entity scanning using recursive VirtualFile traversal.

Two rules are enforced:
- Rule 1: The inverse side of a bidirectional @OneToMany, @OneToOne, or @ManyToMany relationship must declare the mappedBy attribute.
- Rule 2: The inverse side must not carry a @JoinTable annotation.

The owning side is identified by scanning all @Entity-annotated classes in the module (via DiagnosticsUtils.findAnnotatedClassesInModule) for a mirrored back-reference annotation pointing back to the declaring class. For self-mirroring annotations (@OneToOne, @ManyToMany), the target's back-reference must itself carry a non-empty mappedBy to confirm it is the declared inverse side, preventing both sides from flagging each other.

Screen recording:- 

https://github.com/user-attachments/assets/8622535c-d395-4ee8-8cca-7e58f03bb643

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
@@ -13,21 +13,14 @@
 
 package io.openliberty.tools.intellij.lsp4jakarta.lsp4ij;
 
-import com.intellij.openapi.module.Module;
-import com.intellij.openapi.module.ModuleUtilCore;
 import com.intellij.openapi.project.Project;
-import com.intellij.openapi.roots.ModuleRootManager;
-import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.*;
 import com.intellij.psi.search.GlobalSearchScope;
 import com.intellij.psi.tree.IElementType;
 
 import java.beans.Introspector;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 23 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
```diff
@@ -235,4 +235,30 @@ public static boolean isEnumKeyType(PsiType keyType) {
...
+     * Extracts the simple class name from a {@link PsiType}.
+     *
+     * <p>For a plain class type ({@code Department}) returns the class's simple name.
+     * For a parameterized collection type ({@code List<Employee>}) returns the simple
+     * name of the first type argument.
+     *
+     * @param psiType the PSI type to inspect
+     * @return the simple class name, or {@code null} if it cannot be extracted
+     */
+    public static String getElementTypeSimpleName(PsiType psiType) {
```
Rejoice472, line 249: It seems that getElementTypeSimpleName and resolveTargetEntityType share some common logic. Would it make sense to extract the shared functionality into a separate method?

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceBidirectionalDiagnosticsCollector.java
```diff
@@ -0,0 +1,265 @@
...
+     * {@code @OneToMany} ↔ {@code @ManyToOne}; others are self-mirroring.
+     *
+     * @param relAnnotationFQ the fully-qualified annotation on the declaring side
+     * @return the mirrored fully-qualified annotation name
+     */
+    private String getMirroredAnnotation(String relAnnotationFQ) {
+        if (PersistenceConstants.ONE_TO_MANY.equals(relAnnotationFQ)) {
+            return PersistenceConstants.MANY_TO_ONE;
+        }
+        if (PersistenceConstants.MANY_TO_MANY.equals(relAnnotationFQ)) {
```
Rejoice472, line 259: the condition is really needed?

venmanyarun: removed.. fixed in lsp4jakarta as well

src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/DiagnosticsUtils.java
venmanyarun: Fixed.. moved into a separate method

PR #1726, opened 2026-08-27
Add multi-module Maven and Gradle project support

### LINKED ISSUE #1269: Enhance modelling of Liberty modules to include multi-module Liberty projects.
This representation should include:
- the aggregator (`parent`) and submodule (`child`) relationships.
- the root modules, top-level aggregators and independent single module projects
- the leaf modules, e.g. the projects containing Liberty server configuration

This should be kept generic enough so that this model will continue to work with Gradle as-is and with future support for multi-module Gradle projects. See current `LibertyModule` and related classes for reference.

### LINKED ISSUE #1270: Build model containing multi-module relationships from Maven projects.
Scan the project workspace for `pom.xml` files to identify the locations of Maven modules. The Maven Help plug-in can be used to extract the list of submodules from multi-module projects. The results returned from the `help:evaluate` goal appear to be fully resolved values from the effective POM so at least from a functional perspective this is better than reading the `pom.xml` directly with a generic XML parser. With this information, a model (see #1269) can be built and queried to understand the relationship between modules, including which modules are roots and which contain submodules and what those submodules are.

References:
- Maven Help plug-in documentation: https://maven.apache.org/plugins/maven-help-plugin/
- `help:evaluate` documentation: https://maven.apache.org/plugins/maven-help-plugin/evaluate-mojo.html

### LINKED ISSUE #1273: Provide aggregator centric view of projects in the Liberty dashboard.
If you load the finished multi-module Maven project from this Liberty Guide: https://openliberty.io/guides/maven-multimodules.html into LTI 24.0.12 you'll notice that it includes both the top-level aggregator project and the EAR sub-module in the explorer view.

Image

Starting dev mode on the aggregator is equivalent to starting the EAR module. The EAR module requires the aggregator anyways to resolve its dependencies on JAR and WAR. The goal here is to filter the view so that it only includes the top-level Liberty modules and provide a consistent entry point into the projects.

### LINKED ISSUE #1274: Enhance Liberty dashboard view of projects to allow selection of a Liberty module for performing actions.
A multi-module Maven project may contain multiple sub modules that are logical Liberty projects. Imagine an aggregator which includes two EAR projects: `ear1` and `ear2`. The Liberty Maven plug-in requires that you specify which Liberty project is being started when you invoke dev mode.

Start ear1: `mvn liberty:dev -pl ear1 -am`
Start ear2: `mvn liberty:dev -pl ear2 -am`

The user needs to make this choice which requires that the Liberty dashboard be enhanced to accommodate this selection. See mock screenshot below which shows actions for each Liberty sub-module under the top-level aggregator.

Image

This would be a structural change to the tree view. Need to determine whether this is technically possible and the effect it would have on the Gradle view. If this is not possible we will need another solution (e.g. a pop-up which allows the user to select the submodule from a drop-down list) though should be mindful of the user experience.

### LINKED ISSUE #1275: Enhance start actions (liberty:dev and liberty:devc) to include selected modules in the Maven command.
When a Maven multi-module project contains more than one Liberty module, the user needs to select one and the command line for Maven needs to specify that module, for example: `mvn liberty:dev -pl ear1 -am` or `mvn liberty:devc -pl ear1 -am`. Essentially this is appending `-pl <project-selector> -am` to the command. To avoid the complexity of computing relative paths, it may be better to use the artifact ID which uniquely identifies the sub-module. This artifact ID is available from the `pom.xml`. Refer to https://maven.apache.org/guides/mini/guide-multiple-subprojects-4.html#inclusion-and-exclusion for the format of the project selector. Also see: https://github.com/OpenLiberty/ci.maven/blob/main/docs/dev.md for more info on the Maven dev mode command.

With the updates made to the model and UI this work should hopefully be fairly straightforward. See `LibertyDevStartAction` and `LibertyDevStartContainerAction` for the existing code which generates the Maven command.

### LINKED ISSUE #1751: Multi-Module Gradle Task and Directory Targeting Support
Gradle commands for subprojects should be executed from the root project directory using project-qualified task notation (e.g., `:subproject:libertyDev`, `:subproject:libertyDevc`).

Gradle subproject metadata needs to be parsed from `settings.gradle` / `settings.gradle.kts (include(...)` and `rootProject.name)` as well as inter-project dependencies (`project(':name')`).

### LINKED ISSUE #1752: State Lifecycle Tracking & State Indicators (Icons & Tooltips)
In multi-module setups, users need visual feedback showing which applications are running, starting, stopping, or stopped:

- Aggregators and leaf modules need real-time state visualization.
- Aggregator nodes have composite states: if some children are running and some are stopped, a "`mixed/incomplete`" state should be indicated.
- Aggregator tooltips should display summarized statistics (e.g., "`{N}/{total} running`").

### LINKED ISSUE #1753: Implement Terminal Output Watcher for Real-Time State Transitions
The dashboard needed an automated mechanism to transition module states without manual refreshes:

- When dev mode is started, detect when Liberty is ready by listening to the terminal output for the `CWWKF0011I` message to promote state from `STARTING -> RUNNING`.
- Baseline terminal text capture is required to avoid false positives from previous runs.
- When dev mode stops or fails, detect terminal command termination to transition from `STOPPING/STARTING -> STOPPED`.

dessina-devasia:
**Existing Muti-module issues**

Fixes #1269: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5580474332)
Fixes #1270: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5580718696)
Fixes #1273: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5581902359)
Fixes #1274: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5582727605)
Fixes #1275: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5583561065)

**Newly created multi-module issue as a part of task breakdown**

Fixes #1751: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5599259616)
Fixes #1752: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5599506063)
Fixes #1753: [**Fix details**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5599578265)

[**Overall PR Description**](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issuecomment-5599597862)

**Demo video**

https://github.com/user-attachments/assets/e0feef64-74b4-44a6-bc97-e4e54eeabcac


src/main/java/io/openliberty/tools/intellij/actions/LibertyDevStartAction.java
```diff
@@ -13,8 +13,10 @@
 import com.intellij.openapi.vfs.VirtualFile;
 import io.openliberty.tools.intellij.LibertyModule;
 import io.openliberty.tools.intellij.util.*;
+import io.openliberty.tools.intellij.util.LibertyTerminalWatcher;
 import static io.openliberty.tools.intellij.util.Constants.ProjectType.*;
 import static io.openliberty.tools.intellij.util.Constants.*;
+import static io.openliberty.tools.intellij.util.Constants.LIBERTY_GRADLE_DEBUG_PARAM;
```
gilbysunil14: Duplicate import

src/main/java/io/openliberty/tools/intellij/actions/LibertyDevStartAction.java
```diff
@@ -52,10 +54,23 @@ protected void runInTerminal(LibertyModule libertyModule, boolean runInContainer
...
+                        : buildFile;
+                buildSettingsCmd = LibertyMavenUtil.getMavenSettingsCmd(project, settingsBuildFile);
             } else {
-                buildSettingsCmd = LibertyGradleUtil.getGradleSettingsCmd(project, buildFile);
+                // For Gradle child modules, getGradleSettingsCmd needs the root project path.
+                // GradleSettings is keyed by the root project path (parent dir), not the child's.
+                VirtualFile settingsBuildFile = (libertyModule.getParentModule() != null
+                        && libertyModule.getParentModule().getBuildFile() != null)
+                        ? libertyModule.getParentModule().getBuildFile()
+                        : buildFile;
```
gilbysunil14: Same code in the above if block. Extract common code outside the if-else

src/main/java/io/openliberty/tools/intellij/actions/LibertyModuleSelectionDialog.java
```diff
@@ -0,0 +1,123 @@
...
+     */
+    public LibertyModuleSelectionDialog(@NotNull Project project,
+                                        @NotNull String title,
+                                        @NotNull String message,
+                                        @NotNull List<LibertyModule> modules) {
+        super(project, true);
+        this.modules = modules;
+        this.message = message;
+        setTitle(title);
+        setOKButtonText("Select");
```
gilbysunil14: I think we need to externalize this "Select" text

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectMetadata.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.util.List;
+
+/**
+ * Assisted by IBM Bob
+ *
+ * Common contract for build-tool metadata extracted from a project's build file.
+ * Implemented by {@link MavenProjectMetadata} and {@link GradleProjectMetadata}.
+ */
+public interface LibertyProjectMetadata {
```
gilbysunil14: Implementation classes of this interface seems to have similar logic. Would it be better to have an abstract class instead of interface?

src/main/java/io/openliberty/tools/intellij/util/GradleProjectMetadata.java
```diff
@@ -0,0 +1,570 @@
...
+    /** Matches: rootProject.name = 'value' or rootProject.name = "value" */
+    private static final Pattern ROOT_NAME_PATTERN =
+            Pattern.compile("rootProject\\.name\\s*=\\s*[\"']([^\"']+)[\"']");
+
+    /**
+     * Matches the start of an {@code include} statement, capturing everything
+     * after the keyword on the same line.
+     * Handles: include 'a','b'  include('a','b')  include ':a'  include(":a")
+     */
+    private static final Pattern INCLUDE_START_PATTERN =
```
gilbysunil14, line 65: Seems like this pattern will fail if we have some thing like includeBuild('') and treat what is inside this as sub modules.

src/main/java/io/openliberty/tools/intellij/util/GradleProjectMetadata.java
```diff
@@ -0,0 +1,570 @@
...
+    // -------------------------------------------------------------------------
+
+    private void extract() throws Exception {
+        // Project directory: prefer build file's parent, fall back to settings file's parent.
+        Path projectDir = buildFilePath != null
+                ? Paths.get(buildFilePath).getParent()
+                : (settingsFilePath != null ? Paths.get(settingsFilePath).getParent() : null);
+
+        projectName = resolveProjectName(projectDir);
+        subprojects = resolveSubprojects(projectDir);
```
gilbysunil14: Both of these methods seems to be reading the same settings file twice. I/O operations are costly. I would recommend reading a single file once.

src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
```diff
@@ -28,6 +28,51 @@
...
+     * <p>For child modules in a multi-module build the command must run from the
+     * <em>root</em> (parent) project directory so Gradle can locate the settings file
+     * and the full project hierarchy. For standalone projects the module's own
+     * directory is returned.</p>
+     *
+     * @param module The Liberty module to target.
+     * @return The absolute path to the directory from which Gradle must be invoked.
+     */
+    public static String getGradleExecutionDir(io.openliberty.tools.intellij.LibertyModule module) {
+        io.openliberty.tools.intellij.LibertyModule parent = module.getParentModule();
```
gilbysunil14: Add LibertyModule as import. I could find atleast 4 instances in the same file where fully qualified name is used.

src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
```diff
@@ -45,6 +45,49 @@
...
+     * appropriate {@code cd} path is the parent build file's parent directory.</p>
+     *
+     * <p>When {@code module} has no parent (standalone project) an empty string is
+     * returned and the command should be executed from the module's own directory.</p>
+     *
+     * @param module The Liberty module to target.
+     * @return Additional Maven CLI arguments, or an empty string for standalone projects.
+     */
+    public static String getMavenModuleArgs(io.openliberty.tools.intellij.LibertyModule module) {
+        io.openliberty.tools.intellij.LibertyModule parent = module.getParentModule();
```
gilbysunil14: Add LibertyModule as import. I could find atleast 4 instances in the same file where fully qualified name is used. Please review similar instances.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
```diff
@@ -263,6 +268,27 @@ private static ArrayList<BuildFile> getBuildFiles(Project project, Constants.Pro
...
+     *
+     * <p>Maven copies {@code pom.xml} files into {@code target/} during packaging (e.g.
+     * {@code target/m2e-wtp/ear-resources/META-INF/maven/.../pom.xml}). If those copies
+     * are not excluded they appear as phantom top-level Liberty modules.</p>
+     */
+    private static boolean isBuildOutputFile(VirtualFile vFile) {
+        VirtualFile parent = vFile.getParent();
+        while (parent != null) {
+            String name = parent.getName();
+            if ("target".equals(name) || "build".equals(name)) {
```
gilbysunil14: There are similar directories like .intellijPlatform, .idea that can be excluded as well. Make a list and exclude those directories as well.

src/main/java/io/openliberty/tools/intellij/util/LibertyTerminalWatcher.java
```diff
@@ -0,0 +1,195 @@
...
+                        }
+                    }
+                } catch (Exception ex) {
+                    LOGGER.warn("LibertyTerminalWatcher: error reading terminal screen text", ex);
+                }
+            });
+        };
+
+        try {
+            widget.getTerminalTextBuffer().addModelListener(holderRef[0]);
```
gilbysunil14, line 124: A listener is registered here which checks for the start message. What happens if the liberty start is failed? Will this listener get disposed?

gilbysunil14: Another related question, you have mentioned that getText() is a Swing call here. I understand this is for getting the terminal output. Will this call return the currently visible terminal output or the entire output in the terminal? If its currently visible output, is there a chance that we miss finding the start message in case of slow machines and since we are running these operations in EDT?

src/main/java/io/openliberty/tools/intellij/util/LibertyTerminalWatcher.java
```diff
@@ -0,0 +1,195 @@
...
+     * Starts a background poller that watches for the terminal process to finish.
+     * Once {@link ShellTerminalWidget#hasRunningCommands()} returns {@code false},
+     * the module's state is set to {@code STOPPED}.
+     *
+     * <p>Safe to call from any thread.</p>
+     *
+     * @param widget        The terminal widget running the dev-mode process.
+     * @param libertyModule The module whose {@code AppState} to update.
+     */
+    public static void watchForStopped(ShellTerminalWidget widget, LibertyModule libertyModule) {
```
gilbysunil14, line 140: Will this work if we exit from the terminal by pressing ctrl + c? I am not sure if this is a valid question.

src/main/java/io/openliberty/tools/intellij/util/MavenProjectMetadata.java
```diff
@@ -0,0 +1,309 @@
...
+    @Override
+    public List<String> getProjectDependencies() {
+        return projectDependencies != null ? projectDependencies : new ArrayList<>();
+    }
+
+    // -------------------------------------------------------------------------
+    // Parsing
+    // -------------------------------------------------------------------------
+
+    private void parsePomXml(String xmlContent) throws Exception {
```
gilbysunil14, line 63: I can see that you are using a thirdparty library here for parsing pom xml file. Also you were using plain java in case of gradle. I was wondering is there any similar library that we can use there? This is just a suggestion.

gilbysunil14: This looks neat compared to gradle settings parsing ^-^

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
```diff
@@ -246,27 +294,176 @@ public Component getTreeCellRendererComponent(
...
+        }
 
-            return this;
+        /**
+         * Picks the correct state icon
+         * STARTING → starting.svg, STOPPING → stopping.svg.
+         */
+        private static Icon resolveStateIcon(LibertyModuleNode moduleNode) {
+            LibertyModule.AppState state = resolveEffectiveState(moduleNode.getLibertyModule());
+            if (state == null)                         return LibertyPluginIcons.incompleteIcon();
```
gilbysunil14, line 377: Spacing is larger than usual.

src/main/java/io/openliberty/tools/intellij/LibertyModules.java
```diff
@@ -101,10 +109,238 @@ public LibertyModules rescanLibertyModules(Project project) {
...
+                        }
+                    }
+                } catch (Exception e) {
+                    LOGGER.warn("Failed to parse potential parent pom at: " + parentPom.getPath(), e);
+                }
+                continue; // Maven project — no need to check Gradle
+            }
+
+            // Check for Gradle aggregator (settings file in parent dir)
+            java.nio.file.Path parentSettingsFile = GradleProjectMetadata.findSettingsFile(
```
gilbysunil14, line 283: Check for maven and gradle aggregator seems to be identical. Can you extract shared logic?

src/main/java/io/openliberty/tools/intellij/LibertyModules.java
```diff
@@ -101,10 +109,238 @@ public LibertyModules rescanLibertyModules(Project project) {
...
+            if (metadata.isAggregator()) {
+                String parentLocation = module.getBuildFile().getParent() != null
+                        ? module.getBuildFile().getParent().getPath() : null;
+                if (parentLocation == null) continue;
+
+                for (String subprojectPath : metadata.getSubprojects()) {
+                    try {
+                        File parentDir = new File(parentLocation);
+                        File childDir = new File(parentDir, subprojectPath);
+                        String resolvedPath = childDir.getCanonicalPath();
```
gilbysunil14, line 206: File.getCanonicalPath() behaves differently on Windows and macOS because it relies directly on the underlying operating system's file system rules - see Java docs. Should use normalization here.

src/main/resources/icons/state/dark/incomplete.svg
gilbysunil14, line 1: Just curious on where did you get these icons. Will using these icons cause copyright infringment?

gilbysunil14 (commented):
@dessina-devasia Please see my comments.

src/main/java/io/openliberty/tools/intellij/util/GradleProjectMetadata.java
dessina-devasia: Updated regex to match required pattern

src/main/resources/icons/state/dark/incomplete.svg
dessina-devasia: I've used this icons from liberty-tools-eclipse, which won't cause any copyright infringement.

src/main/java/io/openliberty/tools/intellij/util/LibertyTerminalWatcher.java
dessina-devasia: > A listener is registered here which checks for the start message. What happens if the liberty start is failed? Will this listener get disposed?

Great point. To address this, I've updated `LibertyTerminalWatcher` with two layers of cleanup:

- I've registered the listener with `Disposer.register(widget, ...)` so that if the terminal tab/widget is closed by the user at any point, the `TerminalModelListener` is automatically deregistered.

- I've added a background poller that checks `!widget.hasRunningCommands()`. If the server start fails (e.g., port conflict, compilation error, bad config) and the terminal process exits before emitting `CWWKF0011I`, the listener is safely removed and the module state is reset back from `STARTING` to `STOPPED`.

dessina-devasia: > Another related question, you have mentioned that getText() is a Swing call here. I understand this is for getting the terminal output. Will this call return the currently visible terminal output or the entire output in the terminal? If its currently visible output, is there a chance that we miss finding the start message in case of slow machines and since we are running these operations in EDT?

`widget.getText()` queries the underlying `TerminalTextBuffer` (which includes both the scrollback history buffer and the current screen buffer), rather than just the visible viewport window.

Because the full history buffer is queried, even on slower machines where multiple log lines arrive while the EDT queue is busy, the `CWWKF0011I` message remains in the buffer's history and is matched when `contains("CWWKF0011I")` runs. I also updated the buffer comparison logic in `LibertyTerminalWatcher` to ensure that even if buffer scrolling or line wrapping occurs during rapid log output on slow machines, the new output slice safely falls back to inspecting the buffer so the `CWWKF0011I` message is never skipped."

src/main/java/io/openliberty/tools/intellij/util/LibertyTerminalWatcher.java
dessina-devasia: If `Ctrl + C `is pressed during `STARTING` (before the server is up):
Pressing `Ctrl + C` in the terminal sends `SIGINT`, terminating the process. IntelliJ's `widget.hasRunningCommands()` becomes `false`. The background watchdog thread in `watchForRunning` detects that the command is no longer running, removes the listener safely, and transitions the module state to `STOPPED`.

**Normal stop workflow (RUNNING state)**: The intended way to stop Liberty in the tool window is via the Stop action button, which issues q and invokes `watchForStopped` to poll `!widget.hasRunningCommands()` until process termination and flip the state to `STOPPED`.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/util/MavenProjectMetadata.java
dessina-devasia: > I can see that you are using a thirdparty library here for parsing pom xml file. Also you were using plain java in case of gradle. I was wondering is there any similar library that we can use there? This is just a suggestion.

Thanks for the suggestion!

**Maven parsing (`MavenProjectMetadata`)**
For `pom.xml`, we are actually using standard Java JDK classes (`javax.xml.parsers.DocumentBuilder` and `org.w3c.dom`), rather than a third-party library. Since `pom.xml` is standard declarative XML, DOM parsing is built into the JDK, fast, and reliable.

**Gradle parsing (GradleProjectMetadata)**
Gradle files (`build.gradle, settings.gradle, .kts`) are executable Groovy/Kotlin scripts rather than declarative data files.

The official programmatic way to query Gradle is the `Gradle Tooling API`, but that requires spinning up a Gradle Daemon process in the background, downloading distributions, and executing build scripts. This would introduce significant overhead, delay dashboard loading, and require active Gradle daemons.

Therefore, following the implementation in `liberty-tools-eclipse` and `liberty-tools-vscode`, I used lightweight regex/pattern extraction for properties like `rootProject.name, include(...)`, and dependencies. This executes in sub-milliseconds without external dependencies or launching daemon processes.

src/main/java/io/openliberty/tools/intellij/LibertyModules.java
dessina-devasia: Agreed. I have replaced `File.getCanonicalPath()` with `Paths.get(parentLocation, subprojectPath).normalize().toAbsolutePath()` to ensure consistent, cross-platform path resolution across Windows, macOS, and Linux without relying on filesystem-dependent canonicalization.

dessina-devasia:
> @dessina-devasia Please see my comments.

@gilbysunil14 Thanks for the review. I've addressed and replied all the comments. Please have a look.

src/main/java/io/openliberty/tools/intellij/actions/LibertyDevStartAction.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/actions/LibertyDevStartAction.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/actions/LibertyModuleSelectionDialog.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectMetadata.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/util/GradleProjectMetadata.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/util/LibertyGradleUtil.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/util/LibertyMavenUtil.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/LibertyExplorer.java
dessina-devasia: Its addressed. Thank you!

src/main/java/io/openliberty/tools/intellij/LibertyModules.java
dessina-devasia: Its addressed. Thank you!

dessina-devasia:
Fix details of #1269 

1. **Parent-Child & Aggregator-Submodule Relationships**: Modelled via `parentModule` and `childModules` in `LibertyModule.java:72-76`, wired bidirectionally by matching parent declarations and aggregator subproject lists in `LibertyModules.java:183-222`.

2. **Root Modules, Top-Level Aggregators & Independent Projects**: `LibertyExplorer.java:120-125` filters for `getParentModule() == null` to display top-level aggregators and standalone single-module projects at the tree root, creating synthetic aggregators on-demand in `LibertyModules.java:248-275`.

3. **Leaf Modules (Liberty Server Configuration Projects)**: Concrete projects containing Liberty configuration are identified via `LibertyModule.java:157` and receive dev-mode action nodes (`start`, `stop`, `tests`, `reports`) in `LibertyExplorer.java:145,149,266`.

4. **Generic Extensibility for Gradle and Future Support**: Build-agnostic metadata extraction is unified under `AbstractProjectMetadata.java:22`, implemented symmetrically by `MavenProjectMetadata.java:43` and `GradleProjectMetadata.java:45`.

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details of #1270 

Here is the breakdown:

- **Scan workspace for `pom.xml` files:** Handled via IntelliJ's `FilenameIndex` in `LibertyProjectUtil.java:109,237-268`.
- **Build multi-module relationship model:** Handled in `LibertyModules.java:144-222`.
- **Query roots, aggregators, and submodules:** Handled via `LibertyModule.java:157-172` and rendered in `LibertyExplorer.java:118-146`.

**Note:**
The original issue #1270 description suggested `mvn help:evaluate` as an exploratory option. However:
- Running `mvn help:evaluate` on every POM file during workspace scan requires launching multiple external Maven CLI processes, downloading plugins, and creating significant startup latency in the IDE.
- Following the design from **liberty-tools-eclipse** and **liberty-tools-vscode**, I used JDK's built-in XML DOM parser `MavenProjectMetadata.java`. It resolves `<artifactId>`, `<parent>`, packaging, and `<modules>` in sub-milliseconds without external process overhead.

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details of #1273 

In the old version, the flat scanner discovered both `guide-maven-multimodules` (aggregator) and `guide-maven-multimodules-ear` (child module) as separate, independent top-level entries in the explorer view, creating duplicate, confusing entry points.

### Solution

1. **Filtering Out Child Modules from the Top-Level View**:
   In `LibertyExplorer.java:120-122`:
   ```java
   if (libertyModule.getParentModule() != null) {
       continue; // Child modules are skipped at the top level
   }
   ```
   The EAR submodule is no longer shown as a top-level root item; only the top-level aggregator (`guide-maven-multimodules`) is shown as the root node.

2. **Hierarchical Nesting under the Parent Aggregator**:
   In `LibertyExplorer.java:132-146`: The EAR submodule is nested cleanly under `guide-maven-multimodules` in the tree hierarchy.

3. **Multi-Module Command Execution & Dependency Resolution**:
   In `LibertyMavenUtil.java` and `LibertyDevStartAction.java:148-149`:
   - When starting dev mode on the submodule, commands are executed from the top-level parent directory (`getMavenExecutionDir()`) with `-pl :guide-maven-multimodules-ear`.
   - This ensures Maven properly builds and resolves sibling JAR and WAR dependencies before launching Liberty dev mode.

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details of #1274 

Enhanced the Liberty dashboard and action execution to fully support multi-module projects containing multiple Liberty sub-modules (e.g. `EAR 1`, `EAR 2` under a shared aggregator):

* **Hierarchical Tree View**: Aggregators display nested sub-module nodes under the parent project instead of a flat list, with each sub-module receiving its own set of dev-mode action nodes (`Start`, `Stop`, `Run tests`, etc.) `LibertyExplorer.java:132-146`.
* Maven and Gradle actions execute from the root directory with module-specific arguments (`-pl :<module> -am` for Maven; `:<module>:libertyDev` for Gradle) (`LibertyMavenUtil.java:178-195`, `LibertyGradleUtil.java:190-210`).
* Added `LibertyModuleSelectionDialog` as a fallback prompt when actions are triggered directly on an aggregator with multiple child modules `LibertyGeneralAction.java:150-167`
* Extracted `AbstractProjectMetadata` to keep multi-module handling generic and consistent across both Maven and Gradle.

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details of #1275 

Completely handled for both `liberty:dev` and `liberty:devc`.

1. `LibertyMavenUtil.java:64-72` checks if the module is a child module and generates `" -pl :" + module.getName() + " -am"` using the subproject's artifact ID.

2. `LibertyMavenUtil.java:84-90` resolves `getMavenExecutionDir(libertyModule)` to the root aggregator directory so Maven has visibility over the entire project reactor.

3. `LibertyDevStartAction.java:112-118,148-149` appends `moduleArgs` to the start command (`mvn liberty:dev` or `mvn liberty:devc`) and switches directory (`cd "<executionDir>"`) before execution. (Inherited and used directly by `LibertyDevStartContainerAction.java:31-33`)


[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details of #1751 

- `LibertyGradleUtil.java:44-75`: Implemented `getGradleTaskForModule(LibertyModule, String)` and `getGradleExecutionDir(LibertyModule)`.
- `GradleProjectMetadata.java:45-210`: Implemented regex parsing for Groovy & Kotlin DSL `settings.gradle(.kts)` files, supporting multi-module Gradle setups without requiring a Gradle Daemon.
- `LibertyDevStartAction.java:120-140`: Adapted Gradle dev startup to execute qualified tasks from the root Gradle directory.

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details for #1752 

- `LibertyModule.java:46-55`: Defined `AppState` enum (`STARTING`, `RUNNING`, `STOPPING`, `STOPPED`).
- `LibertyExplorer.java:303, 350-410`: Implemented `resolveEffectiveState()`, `resolveStateIcon()`, and `resolveStateTooltip()` to calculate aggregate status and format tooltips.
- `LibertyPluginIcons.java:70-100`: Added SVG state icons (light and dark theme support) for running, starting, stopping, stopped, and mixed/incomplete.
- `LibertyBundles.properties`: Added bundle keys for localized state tooltips (`liberty.dashboard.tooltip.*`).

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Fix details of #1753 

The dashboard needed an automated mechanism to transition module states without manual refreshes:

`LibertyTerminalWatcher.java:1-210`:
- `watchForRunning()` attaches a `TerminalModelListener` to the terminal text buffer and compares against initial baseline text.
- Lifecycle integration with IntelliJ Disposer to unregister listeners upon tab closure.
- Background watchdog polling `!widget.hasRunningCommands()` to handle process crashes or build failures.
- `watchForStopped()` polling process termination for clean shutdown.

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
### Overall PR description

Adds multi-module Maven and Gradle project support 

**Project discovery**
- Parse `pom.xml` and `build.gradle` / `settings.gradle` to extract parent–child relationships and subproject lists
- Build parent <-> child links after every workspace scan
- Filter duplicates from Maven `target/` and Gradle `build/` output directories

**Liberty tool window (dashboard)**
- Child modules nested under their parent node; tree starts collapsed
- Root modules show build-type badge (Maven/Gradle) + state icon
- Child (sub-project) modules show state icon only - no badge
- Per-row state tooltips: *Running / Starting... / Stopping... / Stopped / N/M running

**State icons**
- Five SVG icon pairs (light + dark):
  `stopped`, `running`, `starting`, `stopping`, `incomplete`
- Aggregator state derived from children (any STARTING → STARTING,
  all RUNNING → RUNNING, mixed → incomplete)

**Lifecycle state tracking**

`STOPPED → STARTING → RUNNING → STOPPING → STOPPED`
- **STARTING** set immediately on Start action
- **RUNNING** set when `CWWKF0011I` appears in new terminal output
  (baseline snapshot prevents false triggers from previous runs)
- **STOPPING** set immediately on Stop action
- **STOPPED** set when the process exits

**Other changes**
- Actions on an aggregator node automatically route to its single child,
  or show a module-picker dialog when there are multiple children

[Navigate to Top ^](https://github.com/OpenLiberty/liberty-tools-intellij/pull/1726#issue-5269276254)

dessina-devasia:
Closing this PR and planning to reopen it as a series of smaller, more manageable PRs.

src/main/java/io/openliberty/tools/intellij/util/GradleProjectMetadata.java
gilbysunil14: Could we explore a solution that avoids both reading the file twice and having to iterate over the contents twice? My initial concern was mainly around the repeated file I/O, but storing the entire file contents in a list and doing two separate iterations also adds memory usage and another traversal. If the two operations can be combined into a single pass, that might be more efficient.

src/main/java/io/openliberty/tools/intellij/util/LibertyProjectUtil.java
gilbysunil14: The changes looks good now. Are there any more directories that we can exclude? I can suggest one: "gradle".

PR #1749, opened 2026-09-08
Implements diagnostics for @IdClass primary key class structure

sajeerzeji:
Fixes https://github.com/eclipse-lsp4jakarta/lsp4jakarta/issues/973

A primary key class referenced by @IdClass on an @Entity or @MappedSuperclass must satisfy five structural requirements per the spec. This PR validates each one and reports a targeted diagnostic on the @IdClass annotation when violated:

- IdClassMustBePublic - the referenced class must be public
- IdClassMustHavePublicNoArgConstructor - must have a public no-argument constructor
- IdClassMustBeSerializable - must implement java.io.Serializable
- IdClassMustDeclareEquals - must declare an equals(Object) method
- IdClassMustDeclareHashCode - must declare a hashCode() method

### DIFF
--- a/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
+++ b/src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
@@ -55,14 +55,29 @@ public class PersistenceEntityDiagnosticsCollector extends AbstractDiagnosticsCo
 
                 /* ============ Entity Annotation Diagnostics =========== */
                 PsiAnnotation EntityAnnotation = null;
+                PsiAnnotation mappedSuperclassAnnotation = null;
+                PsiAnnotation idClassAnnotation = null;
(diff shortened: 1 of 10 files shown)


src/main/java/io/openliberty/tools/intellij/lsp4jakarta/lsp4ij/persistence/PersistenceEntityDiagnosticsCollector.java
```diff
@@ -541,4 +556,95 @@ private void validateFieldOrMethodType(PsiJvmModifiersOwner element, PsiJavaFile
...
+
+        boolean hasEquals = false;
+        boolean hasHashCode = false;
+        for (PsiMethod method : idClass.getMethods()) {
+            if (!idClass.equals(method.getContainingClass())) {
+                continue;
+            }
+            if ("equals".equals(method.getName())
+                    && method.getParameterList().getParametersCount() == 1
+                    && "java.lang.Object".equals(method.getParameterList().getParameters()[0].getType().getCanonicalText())) {
```
Rejoice472: better use constant for java.lang.Object

PR #1768, opened 2026-09-16
Enhance LibertyModule model to support multi-module projects

### LINKED ISSUE #1269: Enhance modelling of Liberty modules to include multi-module Liberty projects.

dessina-devasia:
Fixes #1269 

- Added `AppState` enum and multi-module fields (`parentModule`, `childModules`, `buildMetadata`, `appState`) to `LibertyModule.java` along with all necessary accessors
- Introduced `AbstractProjectMetadata.java` - an abstract base class that defines the common contract for metadata extracted from Maven and Gradle build files
- All existing single-module behaviour is unaffected - `parentModule` defaults to `null` and `childModules` defaults to empty

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,7 +51,7 @@ on:
   # To re-trigger after new commits, remove and re-add the label.
   pull_request:
     types: [ labeled ]
-    branches: [ main, 'lsp4jakarta-*-integration*', 'lsp4jakarta-*-release*' ]
+    branches: [ main, 'lsp4jakarta-*-integration*', 'lsp4jakarta-*-release*', intellij-2026.2-test-automation ]
 
 jobs:
   fetch_merge_commit_sha_from_lsp4ij_PR:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -130,11 +130,11 @@ jobs:
         # scheduled workflow) or when specific OS inputs are selected via workflow_dispatch.
         runtime: >-
           ${{ fromJson(
-            github.event_name == 'workflow_call' && inputs.allOS && '["linux","mac","windows"]' ||
-            github.event_name == 'workflow_dispatch' && format('[{0}{1}{2}]',
-              inputs.runLinux  && '"linux",'  || '',
-              inputs.runMac    && '"mac",'    || '',
-              inputs.runWindows && '"windows"' || ''
+            inputs.allOS && '["linux","mac","windows"]' ||
+            (inputs.runLinux || inputs.runMac || inputs.runWindows) && format('[{0}{1}{2}]',
+              inputs.runLinux && '"linux"' || '',
+              inputs.runMac && (inputs.runLinux && ',"mac"' || '"mac"') || '',
+              inputs.runWindows && ((inputs.runLinux || inputs.runMac) && ',"windows"' || '"windows"') || ''
             ) ||
             '["linux"]'
           ) }}
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,7 +51,7 @@ on:
   # To re-trigger after new commits, remove and re-add the label.
   pull_request:
     types: [ labeled ]
-    branches: [ main, 'lsp4jakarta-*-integration*', 'lsp4jakarta-*-release*', intellij-2026.2-test-automation ]
+    branches: [ main, 'lsp4jakarta-*-integration*', 'lsp4jakarta-*-release*' ]
 
 jobs:
   fetch_merge_commit_sha_from_lsp4ij_PR:
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,6 +22,11 @@ on:
         type: string
         required: true
         default: main
+      allOS:
+        description: 'Run tests on all OS (linux, mac, windows). When false, runs linux only.'
+        type: boolean
+        required: false
+        default: false
   workflow_dispatch:
     inputs:
       useLocalPlugin:
@@ -44,6 +49,21 @@ on:
         type: string
         required: false
         default: main
+      runLinux:
+        description: 'Run tests on Linux'
+        type: boolean
+        required: false
+        default: true
+      runMac:
+        description: 'Run tests on macOS'
+        type: boolean
+        required: false
+        default: false
+      runWindows:
+        description: 'Run tests on Windows'
+        type: boolean
+        required: false
+        default: false
   push:
     # Restricted to main to avoid triggering workflows on forks.
     branches: [ main ]
@@ -100,12 +120,24 @@ jobs:
     needs: fetch_merge_commit_sha_from_lsp4ij_PR
     # Run for all test-groups when triggered via workflow_dispatch, workflow_call, or when 'run-ci' label is applied.
     if: github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call' || github.event.label.name == 'run-ci'
-    runs-on: ${{ matrix.os }}
+    runs-on: ${{ matrix.runtime == 'linux' && 'ubuntu-latest' || matrix.runtime == 'mac' && 'macOS-14' || matrix.runtime == 'windows' && 'windows-latest' }}
     name: ${{ matrix.runtime }} - ${{ matrix.test-group }}
     strategy:
       fail-fast: false
       matrix:
-        runtime: [ linux, mac, windows ]
+        # Limit PR/push CI to linux only to reduce the number of GHA jobs consumed by LTI.
+        # The full OS matrix (linux, mac, windows) runs only when allOS=true (set by the weekly
+        # scheduled workflow) or when specific OS inputs are selected via workflow_dispatch.
+        runtime: >-
+          ${{ fromJson(
+            github.event_name == 'workflow_call' && inputs.allOS && '["linux","mac","windows"]' ||
+            github.event_name == 'workflow_dispatch' && format('[{0}{1}{2}]',
+              inputs.runLinux  && '"linux",'  || '',
+              inputs.runMac    && '"mac",'    || '',
+              inputs.runWindows && '"windows"' || ''
+            ) ||
+            '["linux"]'
+          ) }}
         test-group: [ LSP4Jakarta-Unit, Maven-MicroProfile, Gradle-MicroProfile, Maven-Custom-Liberty-Install, Gradle-Custom-Liberty-Install, Maven-MP-Config, Gradle-MP-Config, Maven-MP-SID, Gradle-MP-SID, Maven-NLT-REST, Gradle-NLT-REST, Gradle-Language-Server, Gradle-MP-Language-Server, Gradle-Jakarta-Language-Server, Maven-Multi-Module-MP ]
         exclude:
           # Exclude LSP4Jakarta-Unit from mac and windows - only run on linux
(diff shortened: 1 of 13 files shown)


dessina-devasia:
Test automation is not running while adding `run-ci`, checking the reason

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
```diff
@@ -141,4 +190,62 @@ public ShellTerminalWidget getShellWidget() {
...
+    public void addChildLibertyModule(LibertyModule child) {
+        childModules.add(child);
+    }
+
+    /**
+     * Returns an unmodifiable snapshot of the direct child Liberty modules
+     * registered under this aggregator.
+     */
+    public List<LibertyModule> getChildLibertyModules() {
+        return Collections.unmodifiableList(new ArrayList<>(childModules));
```
vaisakhkannan: @dessina-devasia , noticed a warning in code please check

Image

vaisakhkannan (commented):
Similarly, I noticed that newly introduced methods are marked as unused. Since this is a skeleton PR, that’s expected. Make sure those methods are actually used in follow‑up PRs.

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
```diff
@@ -32,6 +64,23 @@ public class LibertyModule {
     private LibertyRunConfiguration customRunConfig;
     private boolean useCustom;
 
+    // -- Multi-module fields --
+
+    /** The metadata extracted from this module's build file, populated during workspace scan. */
+    private AbstractProjectMetadata buildMetadata;
+
+    /** The parent aggregator module, or {@code null} for standalone / root modules. */
+    private LibertyModule parentModule;
```
mrglavas, line 71: @dessina-devasia This is assuming only one aggregator but I believe a user could define multiple aggregators which include the same module. For example, if Aggregator1 included EAR1 as a sub-module and Aggregator2 included EAR1 as a sub-module, would there be two copies of EAR1 in this model with different parents? Would those copies be kept in sync (e.g. same app state)? I am just trying to understand if the model correctly represents all possible configurations for multi-module projects.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
```diff
@@ -0,0 +1,130 @@
...
+    public String getParentProjectName() {
+        return parentProjectName;
+    }
+
+    /**
+     * Returns the relative paths (or names) of direct child modules declared
+     * by this project.  Never {@code null}; returns an empty list for leaf projects.
+     */
+    public List<String> getSubprojects() {
+        return subprojects != null ? subprojects : new ArrayList<>();
```
mrglavas: How is it possible for `subprojects` to be null? It is already initialized as new ArrayList.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    /**
+     * Returns artifact IDs / project-path segments of inter-project dependencies
+     * declared in this module's build file.  Never {@code null}.
+     */
+    public List<String> getProjectDependencies() {
+        return projectDependencies != null ? projectDependencies : new ArrayList<>();
```
mrglavas: How is it possible for `projectDependencies` to be null? It is already initialized as new ArrayList.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
```diff
@@ -0,0 +1,130 @@
...
+     * own build file.
+     */
+    protected String parentProjectName;
+
+    /**
+     * Relative paths (or names) of the direct child modules declared by this
+     * project (Maven {@code <modules>}, Gradle {@code include(...)}).
+     * Empty list for leaf/standalone projects.
+     */
+    protected List<String> subprojects = new ArrayList<>();
```
mrglavas, line 44: Perhaps make this field final if it is not meant to be overwritten.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Artifact IDs / project-path segments of inter-project dependencies
+     * declared in this module's build file.
+     */
+    protected List<String> projectDependencies = new ArrayList<>();
```
mrglavas, line 50: Perhaps make this field final if it is not meant to be overwritten.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
dessina-devasia: @mrglavas These fields are intentionally non-final - expected subclasses (`MavenProjectMetadata`, `GradleProjectMetadata`) assign them during their parsing phase. 
e.g. After reading `<modules>` from a `pom.xml` or `include(...)` from a `settings.gradle`

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
dessina-devasia: @mrglavas This field is intentionally non-final - expected subclasses assign it during their parsing phase. For example, `MavenProjectMetadata` assigns it after extracting `<dependency>` `artifact IDs` from the `pom.xml`, and `GradleProjectMetadata` assigns it after resolving `project(':name')` references from `build.gradle`.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
dessina-devasia: @mrglavas You're right, `subprojects` is initialized to `new ArrayList<>()` at declaration so it can never be `null`. I've removed the redundant null check.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
mrglavas: @dessina-devasia What would subclasses do differently with creating an empty list? I understand that Maven and Gradle implementations would populate this object differently, but seems like it's the same data structure.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
dessina-devasia: Thank you for pointing that @mrglavas, you're right. `projectDependencies` is initialized to `new ArrayList<>()` at declaration so it can never be `null`. I've removed the redundant null check.

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
dessina-devasia: In the current implementation, each LibertyModule holds a single parentModule reference. This reflects the standard way multi-module projects are structured in both Maven and Gradle:

In Maven, a module's pom.xml declares a single <parent>/<artifactId>, and the aggregator's pom.xml lists child modules under <modules>. A given pom.xml file lives in exactly one directory and belongs to one aggregator.

In Gradle, subprojects are declared via include(...) in a single root settings.gradle. A subproject directory is included by the settings file of its root project.

In the scenario you describe - Aggregator1 and Aggregator2 both including EAR1 - in practice these would be two separate physical directories with two separate build files, resulting in two distinct LibertyModule instances in the registry, each with their own parent and independent app state. They would not share state, which is the correct behaviour.

A scenario where the exact same build file is referenced by two different aggregators is not a valid Maven or Gradle project structure - the build tools themselves don't support it.

That said, I want to make sure the model is correct for all cases you have in mind:
- Are there real-world Maven or Gradle configurations where the same module (same physical build file) could legitimately be included by two different aggregators?
- If yes, should parentModule be modelled as a List<LibertyModule> from the start rather than a single reference?
- Or is the single-parent model the right scope for this phase, with the option to extend later if needed?

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
dessina-devasia: @vaisakhkannan I've resolved the warning

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
mrglavas: In my example, Aggregator1 an Aggregator2 are different compositions that happen to have shared resources. In practice the user builds these as separate projects, but since they have parts in common building Aggregator1 will have built a subset of Aggreagor2's modules. Maven supports this because it only ever sees one reference to EAR1 at a time since Aggregator1 and Aggregator2 are different projects. Do we show the status in the dashboard? Starting EAR1 in Aggreator1 is the same EAR1 in Aggregator2. Perhaps this is a question for the larger group. I am curious if this has been accounted for in the design.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
dessina-devasia: You're right @mrglavas, the subclasses were replacing the field entirely rather than populating the initialized empty list. I've made both fields `final` in this PR. In the next PR (`MavenProjectMetadata`, `GradleProjectMetadata`), the subclass parsing methods will use `subprojects.addAll(...)` and `projectDependencies.addAll(...)` instead of reassignment, which is consistent with the final declaration.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
dessina-devasia: @mrglavas Same fix as above (`subprojects`) applied here - `projectDependencies` is now final. The subclass (`MavenProjectMetadata`) will use `projectDependencies.addAll(...)` instead of reassignment in the next PR.

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
dessina-devasia: @mrglavas In the current implementation, each LibertyModule is keyed by its VirtualFile (the physical build file), so EAR1 would be represented as a single LibertyModule instance regardless of how many aggregators reference it. It can only be linked to one parent in the current model, so if both Aggregator1 and Aggregator2 are open in the same workspace, only the first resolved parent relationship would be retained.

App state is tracked per LibertyModule instance, so starting EAR1 from either aggregator would reflect on the same instance - the state would be consistent.

The limitation is the lost parent link for the second aggregator. This is a constraint of the current single-parent model. 

Could you please let me know if I need to open a separate issue to track this scenario?

mrglavas: @dessina-devasia Yes, open an issue and link it to the epic. This is broader than just implementation. We need to understand what the user experience is. There may be design work for the Liberty dashboard. Check if the other IDEs already address this scenario and discuss with them.

src/main/java/io/openliberty/tools/intellij/util/AbstractProjectMetadata.java
mrglavas: @dessina-devasia Thanks for addressing this.

mrglavas (approved):
@dessina-devasia Thanks for addressing my feedback. Please open an issue for the scenario we discussed.

src/main/java/io/openliberty/tools/intellij/LibertyModule.java
dessina-devasia: Sure @mrglavas. Thank you!

vaisakhkannan (approved):
looks good, Thanks

mrglavas:
@dessina-devasia Though I have approved already, please ensure you run a test build. Thanks.

Issue #1739, opened 2026-09-02
Investigate the test automation support for the inlayhint feature


anusreelakshmi934:
As per the changes in the PR - https://github.com/OpenLiberty/liberty-tools-intellij/pull/1723, we need to run test automation to test these features. Investigate how to do this.

anusreelakshmi934:
Inlay hints are rendered inline inside the editor canvas as virtual text — they are not RemoteText nodes in the accessibility tree that findText() or findAllText() can discover. The existing hoverInAppServerCfgFile / getHoverStringData methods in UIBotTestUtils only work for hover popups (HeavyWeightWindow → DocumentationHintEditorPane or JBHtmlPane). There is no existing helper that extracts inlay hints from the editor canvas.

You cannot use findText() or pixel-coordinate assertions to locate inlay hints, because:
IntelliJ renders them as soft-markup decorations on the editor canvas — they are not accessibility-tree elements.
ProjectFrameFixture.getEditorPane() only matches visible_text from the actual source code, not hint overlays.

anusreelakshmi934:
- The API-level tests are the right layer for inlay hint correctness testing. Tests the LSP layer directly via PropertiesManagerForJava.inlayHint() — fast, reliable, no IDE needed
- UI test via runJs() + InlayModel API - Possible but custom work needed - Would require a new UIBotTestUtils helper that calls editor.getInlayModel().getInlineElementsInRange(...) via remote JS — feasible but fragile/slow
- UI test via visual text matching - Not possible - Inlay hints are canvas overlays, not accessible text nodes in the RemoteRobot component tree

anusreelakshmi934:
As part of the test automation investigation for inlay hints, I looked into what can be tested at the plugin integration test level.
I am planning to add integration tests in MicroProfileConfigJavaInlayHintsTest, modelled after the equivalent tests in intellij-quarkus. The following two settings can be covered for Java files:
- Show converters (BuiltInConverter, StaticMethodConverter)
- Show default values (, defaultValue="PT15M")
I plan to test three scenarios: both enabled, converters only, and default values only — identical to the intellij-quarkus approach.

The Settings UI exposes four inlay hint checkboxes. The remaining two cannot be covered by plugin integration tests:
-  Show Java types
-  Show MicroProfile resolve expressions
[...]

PR #1723, opened 2026-08-21
Add Java inlay hints

anusreelakshmi934:
Fixes #1720 #1721 #1737

Reference PRs:
* https://github.com/redhat-developer/intellij-quarkus/pull/1525
* https://github.com/redhat-developer/intellij-quarkus/pull/1520

Attaching videos and screenshots below:

These checkboxes would be enabled initially by default when opening a new project.

Screenshot 2026-09-02 at 12 06 22 PM

Demo video showing the functions of each check box along with closing and reopening the project - 

https://github.com/user-attachments/assets/ef0d4bdb-a832-4b2a-953c-6b4cbdf12875

- Modified failing test. Added a simple Java Enum @ConfigProperty to the test app. When a property expects an enum type and an invalid value is supplied, LSP4MP provides quick fixes.
-  The initialization options sent to the MP LS are now driven by actual user/project settings (including inlay hints) instead of a fixed hard-coded blob, but validation is explicitly withheld at startup to avoid breaking diagnostics.

REST Endpoint URL CodeLens
image

PR #1773, opened 2026-09-21
Test automation support for the inlayhint feature

anusreelakshmi934:
Fixes #1739 
The following two settings are covered for Java files:
Show converters (BuiltInConverter, StaticMethodConverter)
Show default values (, defaultValue="PT15M")
Three scenarios are tested: both enabled, converters only, and default values only — identical to the intellij-quarkus approach.


anusreelakshmi934:
The builds can be ran and verified once https://github.com/OpenLiberty/liberty-tools-intellij/pull/1723 is merged

